PR #1, opened 2026-07-29
ui: season tabs carry their episode count, and a tick once you are through

GLinnik21:
Parity unit 3 (`docs/parity-gaps.md` §2 theme A / §5c): the season tabs now say how big a season is and whether you have finished it.

## What changed

- **`metadata::Season`** gains `leaf_count` / `viewed_leaf_count` — already parsed at `plex/models.rs:150-153`, previously never copied across — plus **`Season::watched()`**, the season-scope form of the container rule `fetch_detail` applies to a show (`viewed >= leaf && leaf > 0`). It lives on the data so the tab's tick and the coming *Mark Season Watched* row read one truth. The `leaf_count > 0` half is load-bearing: a season with no counts is `0 >= 0`, which would otherwise report as fully watched.
- **`ui/widgets.rs`**: `TabPill` gains a trailing **`TabNote`** — a count, or a tick — one type rung down (`size::CAPTION`), never bold, the pill's own ink one alpha step down. It is an annotation, not a second label. `TabPill::note_w` is the single width source, so pill width, the strip's x-advance and the scroll-into-view geometry cannot disagree.
- **`ui/detail.rs`**: `tabs_layout` returns a `TabLay` (was a 4-tuple) carrying the note; the label centres in the pill *minus* the note so a counted tab keeps label+note centred as one unit.

No new colour literal, no raw font size, no hand-placed text; the note is drawn through `Label` / `icons::draw` (which snaps its own origin).

## Verification

- `make check` — **59/59** (baseline 58, +1: the fully-watched rule, including the no-counts and over-count edges).
- `make` — ARM cross-build green.
- Device, under the TV lock: `./tests/run.py --fps` **5/5** (no playback cases — this is UI-only), then a `DISPLAY` capture of *The Morning Show* (`rk=437`).

Server truth vs. what rendered, from `/library/metadata/437/children`:

| season | leaves | viewed | tab shows |
|---|---|---|---|
| Season 1 | 10 | 10 | ✓ tick |
| Season 2 | 10 | 1 | `10` |

Captured in all four pill states (focused, selected-unfocused, plain, plain-watched): the count is clearly legible and clearly subordinate to the label, and the pill padding stays even on both sides.

https://claude.ai/code/session_018GuSHrq9oydnX7nhKE1Hvc

### DIFF
--- a/rust-modules/src/metadata.rs
+++ b/rust-modules/src/metadata.rs
@@ -77,10 +77,26 @@ pub(crate) struct Episode {
     pub(crate) acodec: String, // Media[0].audioCodec
 }
 
+#[derive(Default)]
 pub(crate) struct Season {
     pub(crate) rk: String,
     pub(crate) index: i64,
     pub(crate) title: String,
+    /// episodes in this season (`leafCount`); 0 when the server sent no count
+    pub(crate) leaf_count: i64,
+    /// how many of those are watched (`viewedLeafCount`)
+    pub(crate) viewed_leaf_count: i64,
+}
+
+impl Season {
+    /// Every episode of this season is watched — the season-scope form of the container rule
+    /// `fetch_detail` applies to a show (`viewed >= leaf && leaf > 0`). It lives here, on the data,
+    /// so the season tab's tick and the coming "Mark Season Watched" row read ONE truth instead of
+    /// each re-deriving the comparison at its own site. The `leaf_count > 0` half is load-bearing:
+    /// a season the server sent no counts for is `0 >= 0`, which would otherwise read as watched.
+    pub(crate) fn watched(&self) -> bool {
+        self.leaf_count > 0 && self.viewed_leaf_count >= self.leaf_count
+    }
 }
 
 pub(crate) struct Related {
@@ -558,6 +574,8 @@ fn fetch_seasons(rk: &str) -> Vec<Season> {
             rk: x.rating_key.clone(),
             index: x.index,
             title: x.title.clone(),
+            leaf_count: x.leaf_count,
+            viewed_leaf_count: x.viewed_leaf_count,
         })
         .collect()
 }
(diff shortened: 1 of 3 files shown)


GLinnik21:
### Review pass — what was found, and what I did with it

A fresh-eyes correctness sweep over the diff (geometry, lifetimes, `Season` construction sites, `TabPill`'s other users) turned up **no lifetime or layout defects**: the pill rect, the `on_axis` cull and `tab_focus_geom`'s scroll extent all agree with the note-inclusive content width, and every `TabNote::Count` pointer is used inside the scope of the `TabLay` that owns its `CString`. `draw_tab_row` and the player Info/Chapters tabs degenerate to the exact previous call. Three items were worth acting on, all in `85cc10c` and **none of them moves a pixel**, so the captures above still stand as evidence:

- **`Season` no longer derives `Default`.** It briefly did, purely so two test fixtures could say `..Default::default()` — which silently absorbs every field added later. Counts are exactly the field whose zero reads as a real value, so the fixtures spell everything again.
- **Documented why the note anchors on the label's *painted* width.** The strip sizes tabs with the **bold** label so an advance can't move with focus; an unbolded segment therefore paints a few px narrower. Anchoring the note on the caller's measured width would pin the note while the label's right edge slid under it — a constant label→note gap traded for a variable one. The gap is the relationship the eye reads, so the painted edge wins; the leftover slack sits inside the pill's own padding and is the same slack the label alone always had.
- **`TabPill::width` now says it covers the label only** — a pill that also carries a note must add `note_w`.

### Two things left open, deliberately

1. **A bare count next to a label that often ends in a number** (`Season 12  12`), and a Plex user trained by the orange badge could read `10` as *unwatched* rather than *total*. The alternative that removes both readings is showing the count on **every** tab and letting the tick sit beside it (`Season 1  10 ✓`) — uniform presence makes "total" the only consistent reading. It is busier, and it needs its own device pass, so it is left for the owner to call rather than slipped in unverified.
2. **`tabs_layout` is uncached on the draw path** — one `CString` + one text measure per season per call, now two with the count. Cheap since `text_width` became a `TTF_SizeUTF8` walk, but a 30-season strip (Top Gear) is the shape that would want `ep_meta_h`'s fingerprinted cache, and **no FPS scene covers it**: `fps:detail-transition` is a movie, which returns from `draw_tabs` before any of this runs. Recorded in the code where the cost lives; worth a follow-up unit (a cache plus a many-season FPS scene) rather than an unverified cache here.

PR #10, opened 2026-07-29
ui: the cast row opens a person page, in the Apple TV shape

GLinnik21:
Unit 17 of the parity-gap work (`docs/parity-gaps.md` §5b). Cast headshots on the detail page focused and animated, and OK did nothing. They now open a person page.

## Why it was unbuildable before

`plex::Tag` parsed three of the six attributes `Role[]`/`Director[]`/`Writer[]` carry and dropped the two that ARE the `personId` — `id` and `tagKey`. Adding all four missing fields makes `GET /library/people/{personId}/media` reachable, which returns everything a person appears in **across every library section in one request**.

## The design: Apple TV, not Plex

Plex's own person page is a field list (left portrait, "Actor, Producer, Director", Born + age, Facebook/Instagram handles, a Filmography button, a count-badged "Movies & Shows in Media Libraries", "Known For", "Featured Videos"). That is the feature checklist. The screen built here is the tvOS shape:

* centred **circular portrait** (260px, requested at 300×300 — the same size the cast circles ask for, so arriving from a headshot is a poster-cache hit, not a fresh `/photo/:/transcode` round trip)
* the **name** at `size::HERO`
* an optional **3-line bio with an inline `MORE`** (`TextView::trailing`)
* plain **`Movies`** and **`Shows`** shelves of ordinary poster cards — `RowStyle::HOME`, the same tiles, springs, unwatched corner angle and amber resume bar every other shelf draws

No full-screen sheets. Only shelves that have content get a heading.

## API notes (all verified live against the PMS on 2026-07-29; now in `docs/pms-api.md` §2c)

* `personId` accepts **either** the numeric `id` **or** the `tagKey` guid — both verified against the same record.
* **Group `/media` rows by each row's own `type`, never by the container's `viewGroup`.** Verified: person 6059's response reads `viewGroup:"movie"` over five `movie` rows *and* one `show`. There is a host test pinning this.
* **`GET /library/people/{id}` is not called at all.** It returns only the tag record — no `summary`, no birth date — and everything in it (`tag`, `thumb`) is already on the `Role[]` row the cast circle was drawn from. So the page opens on the header it was handed and only its shelves are fetched: one request per page.
* `count` is in the spec but **this server does not emit it on `Role[]`** — the DTO treats 0 as "unknown", never "none".
* Headshots keep working through the existing pipeline: the absolute `metadata-static.plex.tv` URL rides `image_transcode_path` as `url=` and the server does the TLS. No new image route.

## How the no-bio header degrades

**PMS has no biography to give, so the bio-less header is the normal case, not a fallback**, and the layout is composed so it reads as finished rather than broken:

* The header is a measured flow — portrait, then `space::XL`, then the name. With no bio the block simply ends after the name and the first shelf follows at the standard block gap. There is no empty box, no "Biography" label with nothing under it, and no placeholder line.
* Portrait and name are both centred on the screen's midline, so the two elements read as one deliberate title block whatever their heights.
* Nothing about the page's proportions depends on the bio: `header_h()` measures what is present, and the shelves below reflow off it. The day someone adds the plex.tv call, the bio slots in under the name and everything below moves down — no layout change needed.
* The one place a bio would show is drawn only when the string is non-empty (`draw_header`), and `TextView` only paints the `MORE` run when the 3-line cap actually hid words.

With a bio-less header the flow lands at: portrait 96–356, name cap band 420–~472, `Movies` heading at 536, poster row 582–957, focused under-title band to 1011 — the first shelf is fully on screen at scroll 0, which is why the scroll target is the shared minimal `reveal` and **not** a lift-to-margin pin (a pin would slide the portrait off the top the instant the page mounted). There is a host test for exactly that.

## Navigation

`Route::Person` in `app.rs` with key / pointer / wheel / press-commit / draw / update arms.

* OK on a cast circle → the person page. Every focusable thing on it is a poster card, so OK takes the same tvOS press (dip on key-down, activate on the spring-back) as the home grid.
* OK on a shelf card → that item's detail page, with the person page left standing behind it (`opened_from_person`), so BACK returns to the same shelf position.
* BACK off the person page → the detail page its cast row was on. `person::leave()` re-opens that item by rk **only** if the page navigated away from it in the meantime; the common case is a free instant return, because `metadata::current()` is still that item.
* The return-to-Person branch is gated on the store still holding a person, so a profile switch (`person::reset()`, wired into `install_pms` next to `browse::reset()`) can't route to an empty screen.

## Async

The store (`rust-modules/src/person.rs`) is the `browse.rs` / `metadata.rs` shape: `task::spawn_small`, a `Mutex` mailbox, a generation atomic, a single-flight flag and a retry backoff — with `close`/`reset` clearing the flags **and** the backoff together with the mailbox (the latch `browse.rs` documents). Two things worth flagging in review:

* `land()` is monotone on the generation, so a slow fetch for the actor you already left cannot clobber a newer result the pump hasn't consumed.
* A **superseded** landing is dropped without arming the backoff — a stale network failure is not news about the person now open, and letting it back off would have delayed the new page's first fetch by ~2s.

## Shared component, not a fork

Rather than paste detail.rs's two-pass shelf loop into a third screen, it moves to **`card_row::strip`**. That is where the culling discipline belongs: it is the loop whose `on_axis` test is the reason only VISIBLE tiles ever call `resolve_tex`, and the 64-slot poster LRU must never see an off-screen request. `detail.rs`'s Related and Cast rows now call it; the only behavioural change is that `strip` gained a `resume` closure (both detail rows pass `|_| None`, as before). Each shelf is also capped at `MAX_ROW_ITEMS`, because past that a `CardRow`'s `scale(i)` clamps to the last cell and a focused tile would never pop at all.

## Tests

`make check`: **67 passed** (+9). New:

* `person.rs` — `/media` rows shelved by their own `type` not the container's `viewGroup`; shelves capped at the spring count and unknown types dropped; a landing from the previous person discarded while still releasing the single-flight; a failed fetch keeping populated shelves and backing off instead of publishing empty; `close`/`reset` clearing the flag and the backoff with the mailbox.
* `ui/person.rs` — focus walking only the shelves that exist; a landing that drops a shelf re-seating the focus and clamping the remembered column; opening the page never scrolling the portrait off the top; the second shelf scrolling fully into view and no further.

The two scroll tests assert on a pure `reveal_block` with the header height bounded from above (the real height measures text, which the host suite cannot link) — so they hold a fortiori for the real layout.

`make` (ARM cross-build): green, no new warnings.

## Device verification: deferred to coordinator

I did not touch the TV. Everything below is outstanding.

**`./tests/run.py --fps` — OUTSTANDING.** This matters for this unit specifically: it is a new poster-shelf screen, and the visible-cells-only `resolve_tex` discipline is exactly what the UI-tier floors catch. Two things to know:

1. I added a **`person-page` UI-tier scene** to `tests/manifest.json` (floor 50, `run_secs` 20, `warmup_s` 6) so the screen is actually covered rather than merely not-regressed. It is driven by a new dev trigger `/tmp/plxnative-detailok`, which presses OK on whatever `detailsec`/`detailcol` focused, without the play path. Scene triggers: `plxnative-detail=2005` + `plxnative-detailsec=1` + `plxnative-detailok`.
2. `detail-transition` is the scene that regression-tests the `card_row::strip` move, since detail's Related and Cast rows now run the shared loop.

`tools/tv-session.sh --screen person=<MOVIE rk>` arms all three triggers in one go (documented in the `tv-session` skill; note the rk is the **movie's**, not the person's — the page has no boot trigger of its own).

**The FIFO drive, one boot.** Target: **ratingKey 2005** — *Wallace & Gromit: The Curse of the Were-Rabbit*. Its cast member **index 0 is Peter Sallis (tag id 6059)**, who is the one person on this server with credits in **both** libraries (5 movies + 1 show), so both shelves populate and the `viewGroup` trap is exercised for real. Idina Menzel (id 161) works too but has only 3 movies, so the Shows shelf would be absent.

Boot with `/tmp/plxnative-detail=2005` and the injected token, then:

```
# 1. detail page for rk 2005 is up (hero focused)
printf 'down '  > /tmp/plxnative-remote   # hero -> "Cast & Crew" (a movie's 2nd section)
# 2. cast circle 0 (Peter Sallis) is focused
printf 'ok '    > /tmp/plxnative-remote   # -> the person page
# 3. EXPECT: circular portrait + "Peter Sallis" centred, a "Movies" shelf of 5 posters
#    below it, and the log line:  person: key=6059 'Peter Sallis' movies=5 shows=1
printf 'down '  > /tmp/plxnative-remote   # Movies -> Shows (scrolls; the Shows shelf is below the fold)
# 4. EXPECT: the "Shows" shelf with 1 poster, Wallace & Gromit's Cracking Contraptions
printf 'back '  > /tmp/plxnative-remote
# 5. EXPECT: back on the rk 2005 detail page, focus still on the cast row,
#    and NO second "detail: rk=2005" fetch in the log (current() was never dropped)
```

Capture `tools/capture-screen.sh /tmp/unit-17.png DISPLAY` at step 3. Judge against §5b: centred circular portrait, name under it, `Movies` shelf of ordinary poster cards — and check the header does not read as a field that failed to load (see "How the no-bio header degrades" above).

Worth also clicking a shelf poster (`ck:X,Y`) and pressing BACK, to confirm it lands back on the **person** page rather than Home.

## Shared-file notes for rebasing

All edits are additive. `plex/models.rs` touches only `Tag`; `ui/detail.rs` touches only the cast OK arm plus the two `draw_strip` → `card_row::strip` call sites and the removal of the now-shared function; `app.rs` adds a route variant and its arms.

https://claude.ai/code/session_018GuSHrq9oydnX7nhKE1Hvc


GLinnik21:
## Follow-up: independent correctness review, findings fixed (4647085)

A separate reviewer swept the diff for navigation bugs, async invariants, panic risk, `static mut` aliasing, hit-test drift and perf. Six findings, all addressed; host suite now **68**.

**HIGH — the return trail lost a level, or pointed at the wrong actor.** `opened_from_person` was set but never cleared where its sibling `opened_from_library` is: not in `home_activate` (whose own comment calls itself "the ONE place" that clears the trail), not in `open_library_card`, and not when routing *into* a person page. Two reachable traces:
* Detail(A) → Person(P1) → shelf OK → Detail(M) → cast OK → Person(P2) → BACK → Detail(M) → BACK → **Home**, skipping Detail(A) entirely.
* Detail(A) → Person(P1) → shelf OK → Detail(M) → Play → OS app-switch (which forces `route = Home` and leaves the flag set) → OK any card → Detail(X) → BACK → **an actor page reached from a different item**.

The `current().is_some()` gate I added earlier only turned the first into a teleport instead of a blank screen; the flag is now cleared in all three places.

**HIGH — the person request was a latch.** `detail::on_ok`'s cast arm raises it, and I polled it next to only two of the four `on_ok` call sites — so `plxnative-detailplay` with focus on the cast row left it set, and it fired on an unrelated OK several screens later. There is now exactly **one** consumer, drained every frame whatever the route, so it cannot latch; `leave()` drops any un-consumed request along with the return rk. Host-tested (`leaving_drops_the_un_consumed_request_and_the_return_rk`).

**HIGH — aliasing UB in `ui::person::update`.** It held a `&'static mut Scene` across `clamp_focus()`, which mints a second one from the same static. `detail.rs::update` carries a comment about precisely this; the pump/clamp pair moved above the borrow.

**MED — the pointer hit-test re-measured the whole flow per tile.** Every `tile_rect` went `child_top` → `header_h` → `text_cap_band`, which is a linear scan of the 160-entry glyph cache — ~48 of those per `SDL_MOUSEMOTION`. Now O(visible) like `library.rs::cell_at`: per-shelf row y computed once outside the loop, column derived arithmetically.

**LOW — two per-frame heap allocations on the draw path**, removed: shelf headings are `c"Movies"`/`c"Shows"` literals, and the name is NUL-terminated once at `open`.

**LOW — a doc that overclaimed.** `FETCHING` bounds spawns per pump; it is not a hard one-worker interlock (a supersede, or a stale landing releasing a newer fetch's claim, can briefly overlap two). Neither can wedge — `land` is monotone and `pump` discards stale generations — but the comment now says what is true rather than implying an invariant a reader might build on. Behaviour deliberately left matching `browse.rs`.

### Came back clean
No reachable panic in the new code (every item access is `.get()`, every fixed-array index comes from `present()` ∈ {0,1}, every `as usize` is `.max(0)`-guarded, shelves capped below `MAX_ROW_ITEMS` so `CardRow::scale` is never fed past its spring array). Hit-test geometry reproduces the draw exactly, including the ScrollColumn translate and per-shelf scroll — the focus pop and press dip are outside the hit rect, matching home/library. Async generation/mailbox/backoff sound in both landing orders; leaving mid-fetch leaks nothing.

Still **not run on the TV** — `./tests/run.py --fps` (incl. the new `person-page` scene) and the FIFO drive remain outstanding, as described in the PR body.

GLinnik21:
## Rebased onto `parity-staging` (tip `ef3ccb0`, 31 ahead of main) — base retargeted

`make check` **143 passed** (staging baseline 132, +11 from this branch), clippy gate clean, `make` cross-builds, no new warnings.

### The one that mattered: the credit index space

Unit 5 folded crew into the same shelf, so `credit(i)` runs **every actor and then every crew member**. My OK arm was written against a shelf of actors only — indexing `cast[i]` at a crew tile silently opens the wrong actor (or, past the end, nobody) while the tile under the focus ring shows a director.

**Decision: a crew tile opens that crew member's page, and that is the behaviour I want.** Crew rows are people with a `tagKey` exactly like actors, `/library/people/{id}/media` answers for a director the same way it does for an actor, and a focusable, animating tile that does nothing is the gap this unit exists to close. So the arm resolves through `credit(col)`, and `crew_credits` now carries `id`/`tag_key` through — which git did **not** flag: `metadata.rs` auto-merged into a `Cast` literal missing my two new fields, a semantic break the compiler caught but the merge did not.

Pinned by a new test, `ok_on_a_crew_tile_opens_that_crew_members_page_not_an_actor_at_the_same_index`, which drives the OK arm at an index past both actors and asserts the store opened the *director's* id — the exact regression a future rebase would reintroduce.

### Also adopted from staging

**`Art::Person` replaces my hand-rolled portrait fallback** (unit 5). Mine was a fork of it and a worse one: it glyphed on an *unresolved* texture as well as an *absent* one, so it would have flashed "no photo" on every portrait for the length of its fetch. The shared art case distinguishes the two.

### Conflicts resolved

- **`plex/models.rs`** — kept staging's improved `thumb` comment (crew arrays carry it too) plus my four `Tag` fields; units 4 and 6 touched different structs, all additive.
- **`ui/CLAUDE.md`** — kept unit 11's **single de-duplicated** `widgets.rs` row (did not reintroduce a second), took my `card_row.rs` row carrying the `strip` clause, kept unit 16's `table.rs` row, kept both the `item_menu.rs` and `person.rs` entries.
- **`app.rs`** — `crate::person::reset()` now sits beside both `browse::reset()` and unit 11's new `pms::reset()`; `home_activate` clears both trail flags **above** unit 11's `status_activate` early return, because a Retry press is still the user acting on Home.
- **`tests/manifest.json`** — `item-menu` and `person-page` are both present (9 UI+player scenes, 21 cases).
- **`tools/tv-session.sh`** — `--screen` line carries both `person=<movie rk>` and `itemmenu`.

`widgets::pill_w` → `Button::pill_w` doesn't touch this branch — the person page sizes no pills.

### The three review fixes from the old `4647085`, re-verified after both rebases

1. **`opened_from_person` trail clearing** — set at 1 site, cleared at 5 (`home_activate`, `open_library_card`, both Detail-BACK arms, and the route-into-Person drain).
2. **Single-consumer drain** — `grep -c take_request app.rs` is exactly **1**. The flag cannot latch.
3. **Aliasing fix** — `pump()`/`clamp_focus()` still run *above* `let sc = scene()` in `ui::person::update`.

Hit-test-matches-draw and the mailbox/generation tests all still pass; the two scroll tests are pure (`reveal_block` with the header height bounded above), so nothing under them moved.

### Device recipe unchanged

**rk 2005**, cast index **0 = Peter Sallis (id 6059)** — still the only person on this server with credits in both libraries (5 movies + 1 show). The FIFO sequence in the PR body is unaffected: for a movie, `down` from the hero still lands on the credits shelf (now titled Cast & Crew and holding crew after the actors), and index 0 is still an actor.

One addition worth a look while you are there: press `right` past the actors onto a **crew** circle and OK it — that is the new behaviour above, and the shelf on rk 2005 has 22 actors before the crew, so `detailcol` is the easier way to reach one headlessly.

### One flake, not mine but worth knowing

`stream::tests::every_failed_open_retires_its_fd_and_leaks_nothing` failed once in ~15 runs and has passed every run since. It counts **process-wide** `/dev/fd` before/after with a `+2` tolerance, so any concurrently-running test that opens a file perturbs it — `crate::log()` opens the event log, and several stores log from their pump. I removed my own contribution (the `leave()` test no longer spawns a detail re-open worker it had no use for), but the underlying race is structural and pre-existing.

PR #23, opened 2026-08-13
detail: a borrowed item says whose copy it is, on the line that already carries the facts

GLinnik21:
Deliverable **E** of the Shared Sources design: on a borrowed item, the detail hero's date/runtime line reads

> 13 Jun 2024 · 1 hr 57 min · **Shared by bamx23**

One more run on a line that is already there — last, after the facts about the item, because it is a fact about where the item came from. It takes that line's own rung and ink (`CAPTION` / `TEXT_TERTIARY`, regular) after the same `.45` middot every separator on that line uses: it is dim because the line is dim.

**On your own server there is no run at all** — not an empty one, not a dangling dot. `Detail::source` is the owner's handle and is empty for our own machine; empty means the flow emits no separator, no run and no draw call, so a single-server install pays nothing for the feature. The first host test asserts exactly that: the borrowed row's event list is the own-server row's with the credit appended, prefix-identical.

The design's rejected alternatives are not implemented and are recorded in the code: a chip in the badges row (those state what the item *is*), its own line in the hero (a fifth line for a fact you read once), a kicker above the title (that is the episode's show name).

### The row became a flow

It was a `dotted_run` + `dotted_run_w` pair — a paint expression and a width expression that had to be kept agreeing — with the yield ladder written in terms of the second. `home::heading_flow`'s doc names that pair as the pattern to stop repeating, so the row now has ONE expression: `facts_flow` hands each run to a closure (the draw paints, the tests measure) and `facts_fit` takes a measure rather than owning one. `dotted_run_w` is deleted; `dotted_run` stays as the identity line's join.

The ladder is unchanged for an item with no source (asserted across every budget, not at a corner): the extent goes entirely, then a date that cannot fit alone is elided. The credit slots in **ahead of both** — the trailing credit yields first, as the design calls it — and monotonically, so a row that dropped its extent can never win the credit back.

One behaviour changed for source-less items, and it is a fix: the fragment's leading separator now depends on whether anything was actually emitted before it rather than on whether the item *had* a date or an extent, so an item whose extent the ladder dropped while its date was empty no longer opens its facts line with an orphan dot.

### Data seam

`metadata::Detail::source` (the owner's handle, `""` for our own server) is `""` at its one construction site, exactly as `pms::HubRow::source` is — nothing new is reachable in the product until the multi-server layer can say which server a fetch was made against (`docs/shared-servers.md` step 2 threads `ServerId` through this struct; the roster field behind it, `Resource::source_title`, is already parsed). It is captured at FETCH time rather than read at paint time, because the page outlives the fetch and the current server can move under it.

### The dev seam (second commit)

`/tmp/plxnative-shared=<handle>` stamps that handle onto every item the session loads. It exists because there is no host runtime and, until the multi-server layer lands, no item on this server can put the run on screen at all — so without it the appearance is unreviewable by anyone. It reads through `dev::read` like every other trigger (so `RELEASE=1` compiles it out) and is deliberately not in `DIAG`.

### Verification

- `make check`: **418 → 422**, run three times before and three times after the dev seam, green each time (4 new tests: absence on your own server, the credit's position/tokens, the bare-line case, and the yield ladder swept over every budget).
- `make` (ARM cross-build): green.
- `./tests/run.py --fps`: 6 passed / 5 failed. Every failure is `only 0 post-warmup loop= samples … scene never entered this screen`, on scenes this change cannot reach — and a concurrent job was driving the same TV throughout (see below). Not attributable to this diff, but **re-run on a quiet TV before merging the stack**.
- **Captures: attempted five times, defeated by TV contention.** Another job in this fleet was arming and clearing `/tmp/plxnative-*` on the same television throughout the window (its `plxnative-detailok` / `plxnative-detailsec` / `plxnative-heroidx` appeared between my arming and my capture; my `plxnative-shared` and `plxnative-token` were deleted under me; the route flipped to `player`, then `login`, then `person`, and two grabs came back `capturedWidth: 0`). The single-TV mutex was held for all five attempts, so the contention was not mine to resolve. **The run has therefore never been photographed** — the one thing still owed on this unit, and the reason the dev trigger above is committed rather than kept local.
- `/code-review` at `high`: no findings in this change. Its six findings are all in the pre-existing `shared-servers-base` commits and are reported to the coordinator rather than fixed here.

### DIFF
--- a/rust-modules/src/metadata.rs
+++ b/rust-modules/src/metadata.rs
@@ -464,6 +464,21 @@ fn convert_ratings(it: &crate::plex::Metadata) -> Vec<Rating> {
 #[derive(Default)]
 pub(crate) struct Detail {
     pub(crate) rk: String,
+    /// Which SERVER this item was fetched from, as the OWNER'S HANDLE ("bamx23") — empty whenever
+    /// it came from the signed-in user's own server, which is every item today.
+    ///
+    /// The person, never the machine: the machine's name (`bx23-ldn`) belongs to the Sources list
+    /// and to a failure read-out, and appears nowhere else in the product. Empty is the ABSENCE of
+    /// an attribution, not an empty one — the detail hero draws no separator and no run at all for
+    /// it ([`crate::ui::detail`]'s facts row), so a single-server library pays nothing for the
+    /// feature: no gap, no dot, no draw call.
+    ///
+    /// Captured at FETCH time and stored, rather than read from `plex::servers::current()` at paint
+    /// time: the page outlives the fetch, and the current server can move under it while a load is
+    /// in flight. Populated by the multi-server data layer when it lands (`docs/shared-servers.md`
+    /// step 2 threads `ServerId` through this struct); the roster field behind it is
+    /// `plex::account::Resource::source_title`.
+    pub(crate) source: String,
     pub(crate) is_show: bool,
     pub(crate) kind: String,       // this item's own type: movie | episode | show | season
     pub(crate) show_title: String, // grandparentTitle — the show name, when this item is an episode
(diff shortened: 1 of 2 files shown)


GLinnik21:

PR #24, opened 2026-08-13
playback: which server an item came from is part of its identity

GLinnik21:
Playback becomes per-server: the server an item came from is captured once, at the request, and carried by value through the whole playback protocol.

## The problem

A `ratingKey` names an item on ONE server. So does a `Part.id`, a `Stream.id`, a `playQueueID`, and the resume point. Today every PMS call on the playback path resolves its server **implicitly, at the instant of the call**, via `plex::client_opt()` — which means *"whichever server the user is browsing right now"*.

With one server that is always the right answer. With a friend's share merged into Home, "the item is from B while current is A" is ordinary, and then the resolve, the PlayQueue, `/decision`, the track PUT, the transcode stop and **every progress report** go to A. The resolve worker alone asked the question **four independent times**, so a switch landing mid-resolve could split one playback across two servers.

## What changed

The server is captured **once**, on the main thread, at `request_play` — the same instant and the same discipline as every other `ResolveEnv` field — and carried by value from there:

```
request_play(sid, …) → ResolveEnv.sid → Plan.sid → CUR_SID  (installed by apply_plan)
```

`apply_plan` stays the sole writer of the route statics and now installs `CUR_SID` in lockstep with `CUR_RK`. Every call in the chain became `client_for(sid)`:

| site | now resolves from |
|---|---|
| `build_stream`'s client bind | `env.sid` |
| `resolve_playqueue` (machine id + `create_play_queue`) | the client passed in |
| `server_decision` / `mde_decision` | the client passed in |
| `put_selection` / `select_streams` | `env.sid` (worker) / `cur_sid()` (in-playback) |
| `transcode_seek`, `retranscode` | `cur_sid()` |
| `scrobble_stop`'s captured client | `cur_sid()`, read on the main thread |
| `report_timeline` | its **argument** — see below |
| `metadata::fetch_playing_item` | `sid` |

`plex::client_opt()` is worth naming as the hazard it is, and `build_stream`'s doc now does: it is not a static a worker could be caught reading, it is a **function call**, so `Send` cannot see it, and a worker that resolves its own server compiles clean and passes every test we have.

## The part that is a review argument, not a test result

**`report_timeline` had to ship in the same commit as `CUR_SID`, and no suite on either tier can prove it.**

The `/:/timeline` reporter runs on a worker every ten seconds for the whole of a playback, and it was the only place in the playback path with **no capture at all** — it re-read the current server on each tick. Split from `CUR_SID`, the app would resolve correctly, play correctly, and quietly write the resume point of a borrowed film onto your own server every ten seconds for as long as it played.

Nothing in either tier would show it:

- the POST is fire-and-forget — its result is discarded, so the app itself cannot notice;
- there is no host runtime, so no host test reaches the live reporter;
- `tests/run.py` grades playback progress from the app's own **heartbeat** (`pos=` on the `loop=` line), not from the server the report reached — so a report landing on the wrong server passes `timeline_climb` and `timeline_post` exactly like a correct one;
- and the harness could not reach a second server at all until the injection landed on this base branch.

So the fix is structural and must be read rather than measured: `report_timeline` takes the `ServerId` as an argument, and `engine.rs` captures it **at the spawn site**, beside the `rk` it already captured by value:

```rust
let (sid, rk) = (crate::route::cur_sid(), crate::route::cur_rk());
…
crate::task::spawn("timeline", move || threads::timeline_thread(sid, rk, st))
```

Two things to check by reading: (1) `cur_sid()` is read on the main thread, after `apply_plan` has installed it and before the worker exists; (2) nothing inside `timeline_thread` or `report_timeline` consults the registry's *current* pointer any more.

The nearest thing to a test is `the_timeline_reaches_the_server_the_item_came_from_not_the_current_one` (below), which grades the routing of one report against two real loopback servers. It cannot grade the **capture** — only that the argument is honoured.

## Machine id: per-server, and the registry outranks the cache

`MACHINE_ID` was "fetched once and never refreshed", and it feeds `uri=server://{machineIdentifier}/…` on the PlayQueue POST. Cached globally, that is server A's fingerprint POSTed to B — naming a server B has never heard of, on a best-effort POST that fails silently.

It is now keyed to the server that taught it (`MACHINE_SID`), and `resolve_playqueue` prefers, in order: the **registry's own per-server id** (`Client::machine_id()`, which cannot belong to another server), then this server's cached one, then `GET /identity`. A server registered with an id skips the round trip entirely.

## Host tests

Three new, and the pure parts only — the rest is the argument above.

- `a_plan_round_trips_the_server_the_request_captured` — graded on the resolve that **fails**, which is the exit `build_stream` takes before any network: a plan that could not resolve still names its server, and `apply_plan` installs it.
- `the_machine_id_cache_is_scoped_to_the_server_that_taught_it` — from both sides: its own server reuses the cache, another server gets nothing and must re-ask.
- `the_timeline_reaches_the_server_the_item_came_from_not_the_current_one` — two real loopback servers (the `stream.rs` idiom). The item plays from B, the user switches to A, the POST must land on B; a closing report to A is the control, so *"A heard nothing"* is a fact about the routing and not about a listener that never worked.

That last one needed `plex`'s registry test seams widened to `pub(crate)` (`register_with_client_id`, `reset_for_test`). The public `register` resolves the device id through `session::load`, which **mints and persists** a uuid — a host test has no business writing one — and the reset is what hands the table back empty, so the next test does not find a `client_opt()` answering `Some` with a client whose port closed when this test returned.

## Two things that moved for a reason

**The Up Next still-warm left `apply_plan` for `pump_play`,** its only caller. `apply_plan`'s contract is to be the sole writer of the route statics; a texture prefetch is not one of them. It is also what lets the host suite reach the install at all — `warm_tex` drags in the poster cache and, through it, a GL call the dev Mac cannot link. Behaviour is identical: one non-test caller, and `UP_NEXT` is installed before either point.

**`ui::detail`'s test `mount` now resets `saved_col`.** It reset the filmstrip sub-row and the season capsules as retained state ("no test may inherit one") but not the per-section column memory that `move_focus` seeds from when it enters a strip — so a test that walked one strip decided where the next test's first DOWN landed. Order-dependent, and therefore invisible until the set of tests changes: it surfaced when `route.rs` gained a case that holds `testlock` across real socket I/O and re-ordered who runs beside whom. `reset_view_state` — the production reset the fixture stands in for — has always cleared it.

## The one hole left, deliberately

`metadata::cached_playing` takes the `ServerId` and does not yet use it: `Detail` carries no server, so the match is still the `rk` alone. It is written up at the call site as a `TODO(shared-servers)` with what it costs once two sources share a shelf (B's playback running on A's stream ids, fps, frame size, markers and chapters — silently) and the one-line shape of the fix (`d.sid == sid &&`). The equality tests inside `metadata.rs` belong to the sibling unit that adds the field; the signature is already threaded so that change is one line.

## Verification

- **Device: `./tests/run.py` → 21 passed, 0 failed, 0 known-gap of 21.** The whole matrix, not just the fps scenes — this touches the playback identity path, so the codec, seek, resume, audio-switch and subtitle cases all apply. `timeline_climb` and `timeline_post` pass on every case that carries them.
- **Host: `make check` green ×3**, plus 12 consecutive runs at `--test-threads=16` (which is how the fixture flake above was found and confirmed fixed). 421 tests.
- **ARM: `make` links clean.**

## Not touched

Below `route::url()` nothing changed — the engine consumes an absolute URL that already carries its host, port and token. `ff.rs`, `stream.rs`, `aq.rs` and the Starfish seam are untouched.

**Known gap for whoever picks it up:** `player::error_shape` still reads `plex::serverinfo::subscription()`, which is a single global refreshed by `install`. For an item borrowed from a friend's server the relevant Plex Pass state is *theirs*, so the "this needs a subscription" read-out can name the wrong server's. It is a diagnostics surface and never a routing input (its own module doc says so), and making `serverinfo` per-server is a `plex/` change outside this unit.

### DIFF
--- a/rust-modules/src/app.rs
+++ b/rust-modules/src/app.rs
@@ -2963,7 +2963,7 @@ pub extern "C" fn plex_run(pms_host: *const c_char, pms_port: c_int) -> c_int {
                         if let Some((part, vc, ac, title, resume_ms, dur_ms)) = leaf {
                             if !part.is_empty() {
                                 log(&format!("plxnative-play: rk={rk} start"));
-                                crate::route::request_play(rk, &part, &vc, &ac, &title, "");
+                                crate::route::request_play(crate::route::surface_sid(), rk, &part, &vc, &ac, &title, "");
                                 let resume = crate::metadata::resume_ns(resume_ms, dur_ms);
                                 let fd = matches!(route, Route::Detail);
                                 start_playback(mt, resume, fd, HUD_HEADLESS_MS, &mut route, &mut played_from_detail, &mut hud_nav);
--- a/rust-modules/src/metadata.rs
+++ b/rust-modules/src/metadata.rs
@@ -894,7 +894,23 @@ pub(crate) fn playing_chapters() -> &'static [Chapter] {
 /// when it IS this item, so playing from a detail page costs no extra GET. Snapshotted into
 /// `ResolveEnv` and handed to the worker; splitting the fetch out lost this and quietly added a
 /// PMS round trip to every play from a detail page.
-pub(crate) fn cached_playing(rk: &str) -> Option<PlayingItem> {
+///
+/// `sid` is the server the item being played came from — the other half of its identity, since a
+/// ratingKey only names an item within one server.
+///
+/// **TODO(shared-servers): the test below is still the rk ALONE, and that is the one hole left in
+/// the playback path's otherwise end-to-end server capture.** It cannot be closed here: `Detail`
+/// does not yet carry the server it was loaded from, and guessing (`sid == current_server()`) would
+/// be a different claim that the real field then has to unpick. What it costs once two sources can
+/// share a shelf: the detail page holds rk `1` from server A, the user starts rk `1` from B's
+/// shelf, this returns A's `PlayingItem`, and the `.or_else(fetch_playing_item)` that would have
+/// gone to B never runs — so B's playback carries A's audio/subtitle `Stream` ids (which
+/// `put_selection` and the timeline report then send to B), A's `video_fps` in the Load esInfo, A's
+/// frame size in the local direct-play gate, and A's markers and chapters. Silently. The fix is a
+/// `sid` on `Detail` and `d.sid == sid &&` on the filter; the parameter is already here so that
+/// change is one line.
+pub(crate) fn cached_playing(sid: crate::plex::ServerId, rk: &str) -> Option<PlayingItem> {
+    let _ = sid; // see the TODO above — the pair test needs a field `Detail` does not have yet
     current().filter(|d| d.rk == rk && !d.audio.is_empty()).map(|d| PlayingItem {
         rk: rk.to_string(),
         audio: d.audio.clone(),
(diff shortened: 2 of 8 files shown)


GLinnik21:

PR #25, opened 2026-08-13
plex: a relay is a 2 Mbit/s tunnel, so the plan changes, not a number

GLinnik21:
The last unit of the shared-servers feature: the **relay bandwidth policy**, and the documentation the rest of the feature reads.

## The policy

Plex's relay is the only connection with a ceiling — a tunnel to a Plex relay host, conventionally `:8443`, capped at 2 Mbps with the server transcoding down to fit. `probe.rs` already ranks it last; what was missing was what playback should do when it wins.

It is a **branch, not a parameter**, because there is no parameter to reach for: `TranscodeSpec` has no bitrate field, `maxVideoBitrate` is a literal on the re-encode branch alone, and on direct play a cap means nothing since no encoder is running to obey it. So `plex::link_policy` denies two things over a relay:

* **direct play** — 31 Mbit/s (the measured share) down a 2 Mbit/s pipe arrives as a stall two minutes in, with nothing on any surface saying why;
* **the container remux** — the half that is easy to miss. It copies the codecs and deliberately sends *no cap*, because a cap is what would force the re-encode it is avoiding. Same bytes, same rate, one layer down.

What survives is the re-encode, the only flavor whose point is that the **server** picks the rate — and the server, on the far end of the tunnel, is the only party that knows the real ceiling.

The tier is now a property of the server (`Client::set_link`/`link()`, lock-free, `LINK_UNKNOWN` until told). `route::build_stream` reads it off the client it already holds — no new global for the resolve worker — and one derived `remux` local keeps the payload guess, the stored flavor a seek rebuilds from, and the `/decision` query in agreement. **route.rs touch: 3 behavioural lines** (`directplay` gate, `let remux = video_dp && link.remux`, and the three `video_dp`→`remux` uses that follow).

**Nothing calls `set_link` yet**, so `link()` is `None` everywhere and the policy is `UNRESTRICTED`: byte-identical behaviour today. "Unknown" is deliberately permissive — assuming relay would downgrade every direct play on earth to a server-side encode to prevent a stall on a connection this codebase has never made.

### Honesty

**No relay connection has ever been observed by this codebase.** `includeRelay=1` has been requested since the first login and every relay connection then discarded unconditionally. The 2 Mbps and the `:8443` are Plex's documentation, not measurement — and this account's share advertises **no relay at all**, so the policy is not merely untested but untestable from here. Said at the policy itself, and in §7 of the design note.

## A regression this work caused, found by review

`includeIPv6=1` on the roster query means plex.tv now offers v6 connections, and the live sign-in chooser takes the **first** `local` match and **persists** it. `stream.rs` parses a dotted quad by hand (`AF_INET`, no DNS), so a v6 literal is undialable: a server listing its LAN v6 first would have signed in "successfully" to an empty Home *and* written that address into the session file, breaking every later boot too. Chooser extracted as `auth::choose_local_connection` (it was not gradeable at all before), guarded on the flag *and* a colon in the address, with the two tests that would have caught it.

## Docs

* **`plex/CLAUDE.md`** gains the registry section it never had — why atomic pointers (`client()` is called three times per key, per visible tile, per frame), why each `Client` is leaked (a re-point costs a worker one request to a stale address, never a dangling pointer), why token generations are globally unique (the poster memo compares one number), plus **granted / pinned / reachable** and what collapsing any two of them causes.
* **`Makefile`** — the surviving half of the self-contradiction: it justified not linking FFmpeg by "SONAMEs move, 55→57→58→59→60", a hundred lines above the rules that build and stage a *pinned* copy. The real reason is different in kind (staged beside the binary, no search path, no rpath — a `DT_NEEDED` would resolve at `exec()` to nothing or to the *television's* FFmpeg). The root `CLAUDE.md` had already been corrected at `97cac1b9`; this was the last place carrying the old model, which is how "just use the TV's FFmpeg for https" keeps coming back.
* **`docs/shared-servers.md`** — §3 separates fixed from still-live (the 8-second trap is exactly as live as it was), §5 marks each step's status, §6 defers to `Shared Sources.dc.html` and names the two things its own earlier draft got wrong (Sources is a library-toolbar chip, the tab strip carries nothing), §7 retires the harness question, §8 shows all four corrections landed, §9 covers all seven units. Measured-facts sections untouched.

## Verification

`make check` **424 passed** ×3 (418 at base) + lint; `make` (ARM) green. **No device run, deliberately:** the documentation half cannot be observed on a television, and the relay half cannot be reached — this account's share advertises no relay connection, so there is nothing to exercise. The TV lock was not taken.

## Findings for other units (from the review, not fixed here)

* `probe.rs:202` — the rank key has no dotted-quad-over-hostname tiebreak, so on the module's own fixture the unresolvable hostname sorts *ahead* of the address measured working. The module doc says the stable order "is not ours to reorder", so this is the probe owner's call; `tests/run.py::pick_connection` already implements the tiebreak.
* `plex/serverinfo.rs:102` (and the same comment at `app.rs:501`) — its correctness argument is "host/port fix at the first install", which the registry deletes; `refresh()`'s single-flight can now let an in-flight fetch for A answer for B.
* `tests/run.py:1660` — the FPS path drops `_skipped`, so scenes skipped for a missing second server vanish from the output and the run exits naming the wrong cause.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -220,7 +220,11 @@ used: **libcurl** (`net.rs`) does the plex.tv account/login TLS+DNS that the raw
   `player/` (buffer-feed engine + worker threads — **`rust-modules/src/player/CLAUDE.md` is the
   playback deep-dive; read it before touching playback**), `ff.rs` (THE demuxer — the **bundled,
   pinned** libavformat shipped beside the binary, *not* the TV's), `stream.rs`/`aq.rs` (HTTP socket
-  → AU pipeline), `net.rs` (libcurl/TLS), and the Plex data layer.
+  → AU pipeline), `net.rs` (libcurl/TLS), and the Plex data layer (`plex/` — **its own
+  `rust-modules/src/plex/CLAUDE.md`**, which the rest of this file never pointed at: read it before
+  adding a PMS query, and before assuming there is one server. There is a REGISTRY behind
+  `client()` now — the app can hold a friend's shared server beside your own, each with its own
+  token, `ratingKey` space and watch state. `docs/shared-servers.md` is the design note).
 - `rust-modules/src/ui/` — **the UI, as a shared design system**: `theme.rs` tokens, the retui core
   (`mod.rs` `Painter`/`View`), reusable components (`widgets.rs`/`table.rs`/`label.rs`/`icons.rs`),
   and the screens (`home.rs`/`detail.rs`/`player_hud.rs`/…). **`rust-modules/src/ui/CLAUDE.md` is the
--- a/Makefile
+++ b/Makefile
@@ -116,11 +116,29 @@ RUST_ENV = RUSTFLAGS="-C target-cpu=cortex-a9 -C target-feature=-neon $(RUST_DEB
 #   libpf-1.0 carries mediapipeline::CustomPipeline (the webOS<11 seek path).
 LIBS_REAL = -lSDL2 -lSDL2_ttf -lGLESv2 -lluna-service2 -lglib-2.0 \
             -lwayland-client -lplayerAPIs -lpf-1.0
-# NOT LISTED, DELIBERATELY: FFmpeg, libcurl and libAcbAPI. Their SONAMEs move between releases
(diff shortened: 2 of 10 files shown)


GLinnik21:

PR #26, opened 2026-08-13
home: the hero says whose film it is, and opens on one of yours

GLinnik21:
Deliverable **C** of the Shared Sources design ruled the conditional in its own list — *"if the hero ever becomes independent of a shelf, it needs E's dim run and the preference comes back."* In this app the hero **is** independent: a rotating billboard built per fetch from Continue Watching plus every recent hub, deduped, capped at 8, turning over on its own 8-second timer. It is nobody's focused tile, the shelf heading beneath it attributes something else, and with a merged pool the front door could open on a stranger's film with nothing on screen saying so.

## What it draws

```
Movie · 2024 · 18 · Shared by bamx23
S1 E4 · Episode title · Shared by bamx23
```

**The rung stays, the ink steps down.** `BODY`, *not* the detail page's `CAPTION` — a rung is a role, and the hero line is BODY because it is read from across the room while E's is CAPTION because it sits under a page of copy. What travels between the two screens is the ink step, secondary → tertiary, after a middot at the separator token's own `.45`. A host test asserts the rung explicitly, because "match E literally" is the plausible wrong answer and it is invisible in a diff.

**The line now has a width budget, which it never had.** This is copy over a photograph, legible only inside the hero's corner wedge — which is gone by four fifths of the frame and leaves the last fifth, the side the composition is usually about, completely alone. `HERO_META_R` ends the line at three fifths, a whole fifth of the panel short of that edge, and every run in the flow is handed a budget running to it. So an over-long handle **elides** where a wrapped block would have spilled: there is no wrap in this flow to reach, only a bound to truncate against, and the line can never become two. The facts ahead of the run keep the hero column exactly as they always have — the handle is last on the line and a fact about where the item came from, so it is the right thing to lose first.

## Own-items-first is an ORDERING, not a filter

`own_items_first` promotes the first owned page to the front and moves nothing else, so `[B,B,O,O,B]` → `[O,B,B,O,B]`: the door opens on your own library and a friend's film arrives one flip in, attributed. Filtering instead would leave a borrowed-only account with **no billboard at all** and would overrule a pin the user made — which is why a pool with nothing of ours in it is left exactly as it is. It needs no switch of its own either: the pool is built from included sources only, so a borrowed hero is always the consequence of a pin.

The pool's entries stop being bare catalog indices and become `HeroSlot` — index plus the handle of its shelf's server. Carried rather than looked up, because this is the one place in the app that lifts items **out of** their shelf order, so an entry that knew only its index would have to scan back to a hub to answer "whose is this" for a fact the build already had in hand.

**With one source none of this exists.** `HubRow::source` is `""` everywhere, so the flow returns before its first run — no pad, no dot, no draw call — and `own_items_first` returns on its first line. Absence, not an empty annotation.

## Two things the flow forced, both real

- `hero_content`'s off-screen cull was measured on the **text column**, and the run is the one thing that draws outside it. Left alone, a flip would take the whole column away while ~400px of that line was still on the panel, and the tail of the meta line would blink out ahead of the item it belongs to.
- `meta_drawn_w` clamps to the column. `TextView::truncates` answers *"were WORDS left unplaced"*, and a single token too wide for the column leaves none — it is ellipsized by a separate safety pass — so the measured string can be wider than anything drawn. Unclamped, the run would start in mid-air past the ellipsis, and an absurd one could hand the flow a negative budget, which `text::elide` reads as "no bound at all".

`SOURCE_PAD` is now one constant for both places this screen states a source: the heading and the hero line are one annotation in two positions, not two spacings.

## Tests

Host **427** (+9: four for the ordering — promotion, no-op, borrowed-only, degenerate ends — and five for the line — absence, composition, the rung/ink step, truncation-not-wrapping at three base widths, and the bound's relation to the wedge, read from `widgets::hero_scrim_a` rather than restated) + lint. `make check` run three times, all green. `make` (ARM) links.

**The two halves have very different verification needs, and that is deliberate.** The ORDERING half is pure list arithmetic and is fully settled here — `own_items_first` is host-tested at every case including the boundary where the owner contributes nothing and a borrowed page may then legitimately lead. The ATTRIBUTION half is typographic and cannot be settled by any host test: nothing below draws a pixel.

**Not verified on device.** No frame of this was ever on a panel; everything below is a recipe, not a result.

### Arm and boot

```sh
tools/tv-session.sh up --screen home                       # (a) baseline: no source
ssh root@<tv> "printf '%s' 'bamx23' > /tmp/plxnative-shared"
tools/tv-session.sh up --screen home --keep                # (b) short handle
ssh root@<tv> "printf '%s' 'a-very-long-plex-account-handle-nobody-would-choose' > /tmp/plxnative-shared"
tools/tv-session.sh up --screen home --keep                # (c) the truncation case
tools/tv-session.sh shot out.png                           # after each; DISPLAY source
```

`/tmp/plxnative-shared=<handle>` is **unit 10's committed trigger, reused — there is no second one**. It stamps one handle onto every item a session loads, which is what a fully-borrowed library looks like, and it stands in only where the real answer is empty, so it can never contradict a source the data layer actually has. `--keep` is required on (b) and (c): `up` glob-clears triggers otherwise. Read once at boot, so it must be armed **before** the launch. One file arms the detail page's run too, which is the point — see the cross-check below.

### Where to look

The hero's **meta line**: the small line directly under the big title/clearLogo and directly above the 3-line synopsis, left column, cap top at **y ≈ 552** for a 3-line synopsis (the column is bottom-anchored on `HERO_TEXT_BOTTOM` 692 and grows up, so a shorter synopsis puts it lower). Not the shelf headings below, and not the title.

### What a CORRECT frame looks like

| # | Frame | Correct |
|---|---|---|
| a | no trigger | `Movie · 2024 · 18` — or `S1 E4 · Episode title`. **Byte for byte today's line**: no trailing dot, no gap, nothing added. This is the whole one-server acceptance criterion. |
| b | `bamx23` | `Movie · 2024 · 18 · Shared by bamx23` on **one line**. The run is the **same size as the facts before it** and visibly **dimmer** than them. The middot before it is dimmer than both. |
| c | long handle | Still **one line**, ending in an **ellipsis** — `… · Shared by a-very-long-plex-acc…` — and ending **well left of the last fifth of the frame** (nothing past x≈1152; the right fifth of the picture stays clear). |

**The three ways this is wrong, in order of likelihood:**

1. **The run is smaller than the facts.** That is unit 10's CAPTION treatment leaking onto this line, and it is the wrong answer here. A rung is a role: the hero line is BODY because it is read from across the room, the detail page's is CAPTION because it sits under a page of copy. **Only the ink step travels.** Correct: same size, one step dimmer (`TEXT_SECONDARY` → `TEXT_TERTIARY`).
2. **The line wraps to two lines** in (c), or the synopsis below it has moved down. It must truncate. A second line here would push the pinned action row into the peeking shelf.
3. **The run runs off into the right of the picture** in (c). It must stop at three fifths of the frame, a full fifth inside the corner wedge's own end — past that there is no scrim under it and it is unreadable over bright artwork.

### Two extra frames worth taking while it is armed

- **Mid-flip.** Wait ~8 s on (b) for the auto-flip and shoot during the slide. The run must travel **with** its column, at the same phase as the title and synopsis. If it blinks out early, or appears on the incoming column before the outgoing one has gone, the cull box or the accessor pairing is wrong (both were bugs found and fixed here, and neither is host-observable).
- **Cross-check against the detail page.** `tools/tv-session.sh up --screen detail=<rk> --keep` with the same trigger armed. The two runs must use the **same words** (`Shared by bamx23` — the person, never a machine name) and differ **only** in rung. Same wording, different size, is correct; different wording is a bug in one of the two.

### What no capture can show

`HubRow::source` is `""` everywhere until the multi-server layer lands, so this is the trigger's behaviour, not the product's. The ordering half in particular is invisible on device with one server — it is settled by the host tests, not by a frame.

### DIFF
--- a/rust-modules/src/pms.rs
+++ b/rust-modules/src/pms.rs
@@ -202,16 +202,63 @@ fn hubs() -> &'static Vec<HubRow> {
 
 // ---- rotating hero pool: curated catalog indices (Continue Watching then Recently Added) ----
 const HERO_MAX: usize = 8;
-static mut HERO_POOL: Vec<usize> = Vec::new();
+
+/// One page of the rotating billboard: a catalog index, plus the handle of the SERVER the shelf it
+/// was drawn from came from ("bamx23") — empty for the signed-in user's own, exactly as
+/// [`HubRow::source`] means it.
+///
+/// The handle is carried on the SLOT rather than looked up from the item's shelf at draw time, and
+/// that is the point of the type existing at all: the pool is the one place in the app that lifts
+/// items OUT of their shelf order ([`own_items_first`] promotes an owned page to the front), so a
+/// pool entry that only knew its catalog index would have to find its way back to a hub through a
+/// range scan to answer "whose is this" — for a fact the build already had in its hand.
+struct HeroSlot {
+    idx: usize,
+    source: String,
+}
+static mut HERO_POOL: Vec<HeroSlot> = Vec::new();
+
+fn pool() -> &'static Vec<HeroSlot> {
+    unsafe { &*std::ptr::addr_of!(HERO_POOL) }
+}
 
 /// number of items in the rotating hero pool
 pub(crate) fn hero_pool_len() -> usize {
-    unsafe { std::ptr::addr_of!(HERO_POOL).as_ref().map(|v| v.len()).unwrap_or(0) }
+    pool().len()
 }
 /// hero-pool item `i`, or None
 pub(crate) fn hero_pool_item(i: usize) -> Option<&'static PmsMovie> {
-    let idx = unsafe { std::ptr::addr_of!(HERO_POOL).as_ref() }.and_then(|v| v.get(i)).copied()?;
-    movie(idx)
+    movie(pool().get(i)?.idx)
+}
+/// Handle of the server hero-pool page `i` came from ("bamx23"), or **empty** for the signed-in
+/// user's own — see [`HeroSlot`]. The hero's meta line draws no run at all for the empty case
+/// (`ui::home::meta_source_flow`), so a single-server library pays nothing for this. Borrowed on the
+/// same terms as [`hub_title`]: main-thread only, valid until the next hub commit.
+pub(crate) fn hero_pool_source(i: usize) -> &'static str {
+    pool().get(i).map(|s| s.source.as_str()).unwrap_or("")
+}
+
+/// **Own items first — an ORDERING, not a filter** (Shared Sources, deliverable C).
+///
+/// A borrowed item may not hold the FIRST rotation while the owner contributes at least one, so the
+/// app's front door opens on your own library and a friend's film arrives one 8-second flip in,
+/// attributed. Everything else about the pool is untouched: it stays merged, in the order the
+/// shelves produced it, and the promoted page is lifted out and re-inserted rather than sorted, so
+/// `[B1, B2, O1, O2, B3]` becomes `[O1, B1, B2, O2, B3]` — one page moves, nothing is dropped and
+/// nothing else is reordered.
+///
+/// Filtering instead would leave a borrowed-only account with **no hero at all**, and would overrule
+/// a pin the user made; that is why a pool with nothing of our own in it is left exactly as it is and
+/// opens on a borrowed page. This is also why the rule needs no switch of its own: the pool is built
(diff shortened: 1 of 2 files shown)


GLinnik21:

PR #27, opened 2026-08-13
detail: Also available — the same film on two servers, as a list you can walk into

GLinnik21:
Deliverable **E** of the Shared Sources design (the *Also available* half; the hero's
"Shared by …" run is #23). When a second pinned source also holds the item on screen, the
detail page's actions row gains an **Also available** pill with a trailing chevron, and it
opens the app's usual `Popover` + `TableView` — never a full-screen sheet:

| row | detail | badge |
|---|---|---|
| `Movies` | `This account · 1 hr 57 min` | `1080p` |
| `LDN Films` | `bamx23 · 1 hr 57 min` | `4K` |

## The design's calls, and where each one lives

**A control in the actions row, not a section of the page.** An item on several servers is
a rare fact; a permanent band would spend a whole row on nothing everywhere else. So the
button is drawn only when a second source holds the film — `alt_sources::is_available()`,
which the row consults before it allocates an index. With one server there is no query, no
control and no layout, the same "with one source, none of this is drawn" rule the shelf
heading's annotation implements as absence rather than as a branch that draws nothing.

The gate counts distinct **sources**, not copies: two copies in two libraries of one server
is not "also available *elsewhere*", and the row naming the second one would lead back to
the machine you are already on.

**The tick marks the copy you are on**, and the badge is the resolution class — the
vocabulary the hero's badges row already speaks, through the same `fmt::resolution`, so one
file cannot be spelled two ways on one screen.

**OK NAVIGATES.** It opens that server's own page for the film rather than swapping the copy
under you — which is also what settles the per-server progress problem: watch state does not
travel between servers, so a design that switched the copy in place would have had to explain
a resume position that jumped.

**Ordering is the copy that plays first**, then quality, then yours before a friend's at equal
quality. The design's own example is the case that separates the first two clauses — a 1080p
copy you are standing on sorts above a friend's 4K one — and that is what the tests pin.

Nothing from the REJECTED list is here: no chip in the badges row, no line of its own in the
hero, no invisible merge, no permanent section, no silent copy switch.

## Identity is the guid, never the ratingKey

Every item-shaped integer in Plex is server-local and starts at 1 — both servers in this
household have a `ratingKey` 4 and a section 1 — so `install()` (the seam the multi-server
data layer feeds) is documented as guid-resolved, and until that lands the store is empty at
every mount.

## What this needed elsewhere

- **`Button` gains a trailing accessory slot.** The leading `icon` is part of the label's own
  statement (the Play triangle *is* "play"); a chevron is a disclosure mark about what the
  control does. `pill_w` grew a `pill_w_full` twin rather than a second sizing path — sizing
  the pill without the chevron and then drawing one pushes the label off-centre.
- **The hero row's focus follows its CONTROL by name** (`HeroCtl`/`hero_set`), not by a `+= 1`
  around one insertion point. That was sound while the restart disc was the only conditional
  control; with a second, two controls can appear in one frame and which of them moved the
  focus depends on where it was sitting. The geometry is one accumulation walk for the same
  reason — the alt pill is a variable-width control in the *middle* of a row that was
  otherwise a pill followed by fixed-width discs.
- **The source switch is performed by `app.rs`, not by the screen.** Re-pointing
  `plex::client()` invalidates every store keyed to the server current when it was filled —
  the hub catalog, the browse grid, an open person page — and none of them carries a server in
  its keys. `install_pms`'s wipe is now `activate_server`, shared by the profile switch and
  this one, and the BACK trail is truncated to its root because `Node::Detail` carries a
  ratingKey and no server (step 2 of `docs/shared-servers.md`).

## Tests

`make check` **432 passed** (from 418), run three times, all green, plus the three named
clippy lints and a `--test-threads=1` pass to catch order dependence. `make` and
`make RELEASE=1` both link.

The decisive call — **OK navigates** — is proven through the real press path rather than left
to a capture: the control's OK, the panel presented, one DOWN, commit; then both ends
asserted. The request names the OTHER server and THAT server's ratingKey, and nothing about
the current page moves (no in-place swap, no second navigation request, same mounted item,
same control set). It is **mutation-checked**: with `request_alt_open` stubbed to a no-op that
test fails and the other 431 pass, so it grades the navigation and not the scenery.

Also pinned: the gate (including the two-copies-on-one-server case, which looks like a working
feature and is not, and a list naming only your own source — which is what "not pinned" means
from here, since the store holds pinned sources and nothing else); the ordering rule, with
input-order independence so it is not luck; that **exactly one** copy is ever marked current —
once when a producer lists the same copy twice, and nowhere at all when the page's own copy is
absent, never twice; that the sort key agrees with the badge drawn beside it; that a copy with
no runtime states only its owner; that OK on the row you are on is not a destination, and BACK
is spent on the panel before the trail.

One order-dependence found and fixed in review: `alt_arm` derived the friend's slot from a
literal, and `plex`'s own tests leave `CURRENT` wherever they last set it, so both copies
collapsed onto one source and the control silently failed to appear when one of those ran
first under the same serial lock.

**Not verified on device.** There is no host runtime, so nothing here has drawn a pixel, and
the fleet device pass was deferred while this unit was queued for the lock. No device result
was collected — nothing below is a claim about how it looked.

What is unverified is appearance and the live press: the pill's width, where the panel lands
(it should flip **above** the actions row — the row sits at y≈812 and the list measures ~224,
so below does not fit inside the bottom keep-out), the badge's inline position in a
`TableView` row, and the OK navigation end to end.

Recipe, executable without me. It reuses **#23's `/tmp/plxnative-shared=<handle>`** — no second
trigger was invented — and needs #23 merged only for the hero's own run; the panel's stand-in
reads the trigger directly, so this works on this branch alone.

```sh
# 1. the row WITHOUT the button — the common case, a film only you have
tools/tv-session.sh up --screen detail=4
tools/tv-session.sh shot without.png

# 2. arm the borrowed-item trigger and relaunch (--keep preserves it while re-arming detail=4)
ssh root@"$(cat .tv-host)" "printf bamx23 > /tmp/plxnative-shared"
tools/tv-session.sh up --screen detail=4 --keep
tools/tv-session.sh shot with.png

# 3. open the panel (focus starts on Play; the alt pill is the control to its right)
tools/tv-session.sh key right
tools/tv-session.sh key ok
tools/tv-session.sh shot panel.png

# 4. the decisive one: commit the OTHER copy
tools/tv-session.sh key down
tools/tv-session.sh key ok
tools/tv-session.sh shot destination.png
tools/tv-session.sh log 'altsources|route='
```

A CORRECT set of frames:

- **without.png** — the actions row is `[Play] (✓)` exactly as today. No third control, and no
  gap where one would go. If rk=4 carries a resume point the row is `[Resume] (↺) (✓)`; that is
  fine, the point is that there is no *Also available*.
- **with.png** — one more control, *between* the play group and the watched disc:
  `[Play] [Also available ⌄] (✓)`. The chevron is the same mark the Library's toolbar chips
  carry. The hero is otherwise unchanged — no new band, no chip in the badges row.
- **panel.png** — a rounded panel ~700 wide hanging off the button's left edge and sitting
  ABOVE it, over a light scrim (0.45, the Library chip menus'). Two rows: `Movies` /
  `This account · 1 hr 57 min` / `1080p`, ticked and highlighted; `LDN Films`… no — the
  stand-in labels both rows with your own library title, so expect the SAME label on both, the
  sub-lines differing (`This account` vs `bamx23`) and the badges differing (`1080p` then
  `4K`). **The 1080p row must be FIRST and ticked** — that is the ordering rule on screen: the
  copy that plays outranks the better copy.
- **destination.png** — a DIFFERENT page: the route dipped and came back. The film is the same
  (the stand-in's second slot is the same machine, so its ratingKey is honestly the same film),
  and the panel is gone. The log must carry `altsources: source switched to slot N for rk=…`.
  **If the frame shows the same page with different content, that is the rejected design and a
  real bug** — but note the stand-in cannot produce a genuinely different resume position,
  because both slots address one server.
- The event log should also show `altsources: stand-in for rk=4 on slot N (dev)` once per page,
  never once per frame.

What the stand-in cannot stand in for, so do not read these into a frame: a different library
name, a genuinely different file, a per-server resume position, or a source going offline.


GLinnik21:

PR #28, opened 2026-08-13
plex: artwork and server facts belong to a server, not to the process

GLinnik21:
## The bug, measured on the wire before a line was written

From the dev Mac, against our own server and the friend's share, each with its own token:

```
GET /photo/:/transcode?width=250&height=375&minSize=1&url=/library/metadata/1/thumb/1778526065
  -> our server:     Frozen                25658 B   sha a211b9ac5a
  -> shared server:  Призраки в Венеции    11707 B   sha 59928cd5b1
```

One request path, two different films — and rating keys 2, 3 and 4 collide the same way. Rating
keys are server-local integers from 1 on every Plex server, and the thumb path's trailing
timestamp is a cache-buster the other server ignores, so these paths collide **by construction**,
not by luck.

The poster store keyed a slot on that path alone, and its workers fetched every poster from
`plex::client()` — the *current* server, whatever item the slot belonged to. On a merged Home
that is the wrong picture on a card, filled by a request to the wrong host carrying the wrong
grant.

## What changed

**`posters.rs` — a `ServerId` rides every entry point.** It is a small field on `Pslot`, not a key
prefix (keys already run ~135 B into a fixed 256-byte array — measured at 133 B and 144 B against
the real share, so a prefix would eat real headroom), and `slot_matches` compares the `u16` first:
the half that can reject without touching the array, on a path walked for every slot of every
probe of every visible tile. `lookup` writes the slot's server on the **main thread** when it
claims it; the worker resolves `client_for(that id)` and never asks what the current server is —
the question that has no correct answer on a worker.

**The key memo** had no host and no token in its key and one token generation for the whole map.
The server is now part of the key (so server B's card can never be built from A's memoised,
token-bearing path — a request to B carrying A's grant is a 401) and the generation is part of
each **value**, because with a table of servers there is no single "the token" left to compare
against.

**Layering.** The store called `crate::stream` directly, behind the back of the layer whose module
doc claims to be the only code that touches the socket. It now goes through the new
`Client::fetch_built` — deliberately not `get_bytes`, which would append a second `X-Plex-Token`:
the poster path *is* the store key and already ends in one.

**`serverinfo.rs` — per server, keyed the way the registry is.** Version, subscription **and the
in-flight latch** are arrays indexed by the registry's own slot number, bounded by the registry's
own `MAX_SERVERS` so a second constant cannot drift out of agreement with it. The latch mattered
as much as the values: a global one silently skips the second server's fetch while the first is in
flight, and the answer it never fetched is then read as fact. An id that names nothing reads
`Unknown` — never a neighbour's answer, never a panic. `register` (not `install`) now does the
refresh, because `install` only ever reached the current server: a shared server registered beside
it would have stayed permanently `Unknown`, and `player::error_shape` blames a missing Plex Pass
only on a *known-free* server — the one sentence you least want aimed at the wrong owner.

**Call sites.** The three UI resolvers keep their signatures and gain `_on(ServerId, …)` twins; the
bare form is exactly `_on(current_server(), …)`. A one-server install is byte-identical, and the
item-server plumbing has a door to walk through. One behaviour change falls out for free: a poster
resolve before `plex::install` used to panic through `client()` and now degrades to no artwork.

## Verification

- **`make check` × 3, all green: 421 passed** each run (Makefile nightly + the three named clippy
  lints).
- **`make`** — ARM cross-link green, `pkg/plxnative` is an 8161720-byte ARM EABI5 ELF, md5
  `542779b22eaf048b07f861655bae30ed` reproducible across a clean relink.
- **`tools/fwcompat.py`** — unchanged matrix, OK on 4.4.2 → 11.2.0 (no linkage or FFI change).
- `cargo +nightly build --lib --no-default-features` (the RELEASE feature set) clean.
- **`code-review` skill**: no findings in this unit's files. Three findings landed on other units'
  files — relayed below, deliberately not fixed here.
- **Not verified on device.** The fleet's TV mutex never came free for this unit, so nothing in
  this PR has been seen on a television. No contended screenshot is offered as evidence.

Everything below is executable without me. **Two parts: the first works on any merged stack today;
the second is the one that actually exercises this change.**

### Part 1 — the regression check (one server, ~3 minutes)

```sh
tools/tv-session.sh up --screen home
tools/tv-session.sh shot /tmp/home.png
tools/tv-session.sh up --screen library        # first library tab = Movies on the dev server
tools/tv-session.sh shot /tmp/library.png
tools/tv-session.sh log 'pms: server'
```

**What a correct frame looks like.** Every poster tile carries real artwork — a photographic film
poster — and not the dark grey-to-black gradient skeleton the store draws while a fetch is
pending or failed. This unit rewired *how every one of those images is fetched*, so the failure
mode it would produce is loud, not subtle: a screen of grey skeletons where the posters were.

- **Home**: the full-bleed hero backdrop behind the title at the top, the film's LOGOTYPE (not
  plain text) over it, and the poster shelves below it, all with artwork.
- **Library grid**: a 6-across grid of posters. On the dev server's Movies section, sorted by
  title, row 1 reads: **Frozen · Frozen II · Hannah Montana 20th Anniversary Special · The Hobbit:
  An Unexpected Journey · The Hobbit: The Desolation of Smaug · Home Alone**. Only the *focused*
  tile carries a title label under it; the rest are bare posters, so judge them by the artwork.

**The sharpest and cheapest check is the log line**, and it doubles as proof the commit landed:

```
pms: server 0 version=1.43.3.10861-… plexPass=true      <- must appear (note the slot number)
pms: version=…                                          <- must NOT appear (the old global form)
```

`tools/tv-session.sh log 'pms: version='` finding nothing is the assertion; the slot number is
this change.

### Part 2 — the two-server proof (the frame that matters)

**On a one-server install this unit is deliberately invisible** — byte-identical behaviour — so
Part 1 cannot fail in an interesting way. The interesting frame needs a second server registered.

The credentials come from the friend's real share, `bx23-ldn` (owner `bamx23`), reachable at
`<share-ip-redacted>:<share-port-redacted>` over plain HTTP, machine id `<share-machine-id-redacted>`.
Add this block to the gitignored `tests/manifest.local.json` (its access token is resolved from
plex.tv at run time, exactly as the harness already does — nothing is stored):

```json
"shared_server": {
  "name": "bx23-ldn",
  "machine_id": "<share-machine-id-redacted>",
  "host": "<share-ip-redacted>",
  "port": <share-port-redacted>
}
```

`./tests/run.py --fps --list` then prints `second server: bx23-ldn … configured`, and
`./tests/run.py --shared-server …` writes `/tmp/plxnative-servers` for the run.

**2a — if a sibling unit's ingest landed** (anything that calls `plex::register` for the
`/tmp/plxnative-servers` entries), boot Home with that file armed and read the log:

```
plex: server slot 1 registered at <share-ip-redacted>:<share-port-redacted>
pms: server 0 version=1.43.3.10861-… plexPass=true
pms: server 1 version=1.43.3.10861-07dfddaeb plexPass=…      <- THE serverinfo half of this PR
```

**Two `pms: server N` lines is the whole point.** Before this change there was one global
in-flight latch, so the second server's fetch was *skipped* while the first was running — and its
answer never arrived, while `subscription()` went on answering confidently on its behalf.

Then look at the shelf headed **"Recently Added in LDN Films · bamx23"** (the shared server's
shelf; the heading annotation landed in `4b8a1034`). Its tiles must show **that** library's
posters — a Russian-titled catalogue: *Призраки в Венеции*, *Всё везде и сразу*, *Аватар: Путь
воды*. If those tiles instead show **our** films (*Frozen*, *Home Alone*, *Frozen II*) or grey
skeletons, that is precisely the pre-fix bug on screen.

**2b — if no ingest landed**, apply this 20-line throwaway patch (compile-checked here, never to
be committed). It registers the injected servers and points rating keys 1–4 — which exist on both
servers as four different films — at the shared one, using our own server's paths:

<details><summary>experiment.patch</summary>

```diff
--- a/rust-modules/src/app.rs
+++ b/rust-modules/src/app.rs
@@ -492,6 +492,12 @@
                 log(&format!("servers: {} extra server(s) injected, {usable} usable", v.len()));
+                // EXPERIMENT (not for commit): register the injected servers so the poster store
+                // and serverinfo can be observed with TWO servers live.
+                for s in v.iter().filter(|s| s.usable()) {
+                    let id = crate::plex::register(&s.machine_id, &s.host, s.port as i32, &s.token);
+                    log(&format!("servers: registered {} as slot {}", s.name, id.raw()));
+                }
             }
             Ok(_) => {}
         }
--- a/rust-modules/src/ui/widgets.rs
+++ b/rust-modules/src/ui/widgets.rs
@@ -102,7 +102,24 @@
         Art::Poster(m) => {
-            let t = m.map(|m| resolve_tex(&m.thumb, 250, 375, 0)).unwrap_or(0);
+            let t = m
+                .map(|m| {
+                    let rk = m.rk.parse::<i64>().unwrap_or(0);
+                    let srv = if (1..=4).contains(&rk) {
+                        crate::plex::ServerId::from_raw(1)
+                    } else {
+                        crate::plex::current_server()
+                    };
+                    resolve_tex_on(srv, &m.thumb, 250, 375, 0)
+                })
+                .unwrap_or(0);
             if t != 0 {
```

</details>

Then `make deploy` with `/tmp/plxnative-servers` armed and boot `--screen library`. **The correct
frame, tile by tile** (6-across, Movies sorted by title):

| position | normally shows | must now show |
|---|---|---|
| row 1, col 1 | Frozen (ice-blue) | **Призраки в Венеции** — dark green/blue Poirot poster |
| row 1, col 2 | Frozen II | **Аватар: Путь воды** — blue underwater poster |
| row 1, col 6 | Home Alone | **Всё везде и сразу** — the multiverse collage |
| row 3, col 1 | The Substance | **Гарри Поттер и Тайная комната** |
| every other tile | its own poster | unchanged |

**This experiment is a no-op under the old code and a visible swap under the new one** — that is
the property being tested. Before the fix, all four tiles resolve to the identical path in the
identical slot and the grid looks entirely ordinary; after it, the server is part of the slot's
identity, so four tiles come from the other library.

Afterwards: `git checkout -- rust-modules/src/app.rs rust-modules/src/ui/widgets.rs`, rebuild, and
`tools/tv-session.sh down` — which strips the injected token **and** the second server's token
from `/tmp`.

### FPS gate

This is an artwork change on a per-frame path (`poster_key` runs three times per key per visible
tile per frame), so the scenes worth running are the UI tier: `./tests/run.py --fps`, in
particular `fps:home-grid`, `fps:library-scroll` and `fps:home-idle`. The hot path got *cheaper*
(one atomic acquire load per key where there were three relaxed+acquire pairs), so a regression
here would be a surprise worth chasing.

## Findings outside this unit (relayed, not fixed here)

The review pass turned up three defects on the base branch, all in other units' files — fixing
them from here would collide with those units:

1. **`account.rs:106` + `auth.rs:396` (high).** `resources()` now asks for `includeIPv6=1`, but the
   live selector still takes the first `local && !relay` connection with no address-family filter.
   plex.tv can list a LAN IPv6 connection first; `stream.rs` parses dotted quads only, so that
   address gets persisted into the session and every PMS request fails on every later boot.
2. **`probe.rs:202` (medium).** The candidate sort has no "dotted quad before hostname" tiebreak,
   so for the real share the first candidate is the owner's internal hostname, which does not
   resolve from here — the harness twin (`tests/run.py::pick_connection`) implements exactly that
   tiebreak and documents why.
3. **`servers.rs:143` (low).** `install` passes an empty machine id, so `same_server` falls back to
   address equality; a server that changes address gets a *second* slot, splitting every per-slot
   store — including the two arrays this PR adds. `session.server.machine_id` is available at both
   `install` call sites.


GLinnik21:

PR #29, opened 2026-08-13
auth: discover every server the account can reach, and persist the roster

GLinnik21:
Unit: **Discovery & roster** — the gate on the whole shared-servers feature. If this picks the wrong
connection, no other unit's work is reachable however correct it is.

## What changed

`auth.rs`'s `discover_and_store` kept exactly ONE server and threw the rest of the account away:
both of its passes filtered on `c.local && !c.relay`. Against a real share that is worse than
useless — `Connection.local` means "this address is RFC1918", not "you are on that LAN" — so it
selected the *owner's* `<owner-lan-ip>`: 8 s of timeout from here, and the worse outcome is that it
succeeds, against whatever else sits at that address on our own LAN.

It is now real discovery:

- **Every** resource that `provides` a server, ours and each share, is turned into ranked candidates
  by `plex::probe` and dialled in order.
- The probe is `GET /identity`, **unauthenticated on purpose** and with the token deliberately not
  sent: a probe can land on a machine that is not the one we asked for, and that machine must never
  be handed a live credential.
- The answer is **verified** — its `machineIdentifier` must equal the resource's `clientIdentifier`
  before the connection is accepted. Anything else is discarded and the next address tried.
- **401 is its own state**, never "unreachable" — but it ends the **candidate**, not the server.
  Because the probe is unauthenticated, a 401 is not that server refusing our credential; it is
  whatever is listening refusing an anonymous request, and that need not be the server at all. Only
  a clean sweep — every dialable address refusing — is reported as a refusal.
- **BACK during discovery stops it.** Dialling holds sockets (2 s a handshake, 15 s on a peer that
  connects then says nothing), so the worker can outlive the screen. `cancel` is only a phase flip,
  so discovery polls for it between servers and again immediately before it touches anything global
  — registering would retarget `client()` out from under the Home the user had gone back to, and the
  session write would clobber the session `cancel` had just restored.
- Everything that answered is `register`ed with its **real machine id** and its **own**
  per-(user, server) token. Our own server stays `current`.
- The roster is persisted in `session.rs` as `sources` (+ a `pinned` list for Home), beside the
  primary `ServerRef`, which is written exactly as before.

Single-server behaviour is unchanged by construction: the same `ServerRef` is written, plus a
one-entry roster nothing is obliged to read.

## Notable decisions

- **Sequential probing, not parallel.** `stream.rs` bounds a handshake at 2 s, the dialable list is
  short (policy has already dropped the owner's LAN address, and this client cannot speak to the
  https/hostname candidates), and it runs on the login worker behind a spinner. Racing would buy a
  couple of seconds for N threads that each have to capture at the spawn site.
- **Discovery does not, and cannot, prove the token works** — `/identity` answers 200 to anybody. A
  grant revoked since the `/resources` fetch still probes as reachable and surfaces on the first
  authenticated request. That limit is stated where the probe path is defined rather than papered
  over; an authenticated second probe was considered and rejected because a transient failure
  between two requests would then drop a good server from the roster.
- **The failure copy is now per outcome, not one line.** "No local Plex server found on this network."
  was the old *policy* talking — a server reached over the internet was a failure by construction,
  so the message named the LAN.
- **A profile switch re-keys the whole roster** from the `/resources` response it already fetches
  (`accessToken` is per (user, server), so a share left on the previous profile's token 401s
  everything). Computed before the switch is known to have worked, applied only on success —
  persisting a roster re-keyed for a profile we did not switch to would delete the user's shares
  from disk.

## The two items the coordinator forwarded

**1. Unit 14's IPv6 fix (`choose_local_connection`, PR #25).** This PR deletes that chooser — it is
the legacy selector being replaced — and makes its property **structural instead of a filter that
can be forgotten**: nothing but a *dialable* candidate (plain HTTP to four decimal octets, all
`stream.rs` can build a `sockaddr_in` from) is ever dialled, and only a candidate that **answered**
becomes a stored source. An address can therefore only be persisted after it has been connected to.
Carried over with its own test — `only_an_address_this_transport_can_dial_is_ever_chosen_and_stored`
— including unit 14's sharpest case: an address bearing colons whose `IPv6` flag lies and says
`false`.

*Merge note:* if #25 lands first, resolve by **deleting `choose_local_connection` and its two
tests**; the property they pin is asserted by the test named above.

**2. `probe.rs` had no numeric-vs-hostname tiebreak.** Added as the fourth ranking axis: inside a
tier, a numeric literal outranks a **hostname**. The transport has no resolver at all, and this
account's share lists an internal name that does not resolve ahead of the public address that
answers in 115 ms. Ranked, not dropped — a name is the only thing TLS can ever validate.
Test: `a_hostname_ranks_behind_an_address_that_can_actually_be_dialled`.

## Also hardened while the file was open

Every struct in `session.rs` takes `#[serde(default)]` on the **container**, and the three list
fields decode element by element. `#[serde(default)]` covers a field that is ABSENT, not one present
and the wrong shape — and for this file the cost of that difference is not an empty roster, it is a
`Session` that will not parse: a silent sign-out and a QR code to scan on every boot. `ServerRef`
had no per-field defaults at all, so a session file missing one field already failed the whole parse.

## Tests

`make check` green **3/3 runs**, 433 passing (was 428 at the base). `make` (ARM cross-build) links.
`code-review` run **twice**; every finding in this unit's files fixed. From pass 1: a roster wiped
by a *failed* profile switch, and a plex.tv fetch failure reported as "this account has no server".
From pass 2, both of which changed behaviour rather than wording:

- **BACK during discovery could clobber the resumed session** — the cancel guard was checked before
  a step that used to be a single HTTPS fetch and now holds sockets for seconds, and everything after
  it is global (`register`, `set_current`, the session write).
- **A single 401 abandoned the whole server**, skipping its other advertised addresses. The probe is
  unauthenticated, so that 401 is not evidence about the server's credential at all — DHCP moves the
  PMS, a NAS answers 401 at the old address, and the user's own server is dropped without the second
  address ever being tried.

Also from pass 2: the `Refused` copy said "your Plex server" for a refusal any server can raise, and
the profile-switch token fallback took the *first* server in the response when it had no machine id
to match on — which now hands a share's token to the primary, since plex.tv lists shares first.

New host tests, all pure:

| test | pins |
|---|---|
| `a_sign_in_to_a_two_server_account_settles_on_one_address_each_ours_first` | the whole of discovery over the measured account, socket faked |
| `the_three_empty_outcomes_are_distinguished` | NoServers / Silent / Refused, and a share-only sign-in |
| `a_response_from_the_wrong_machine_is_rejected_and_the_next_address_is_tried` | identity verified before acceptance |
| `a_401_moves_to_the_next_address_and_only_a_clean_sweep_is_a_refusal` | 401 is its own state, and costs one address not the server |
| `a_flow_that_moved_on_mid_dial_registers_nothing_and_stores_nothing` | BACK during discovery mutates no global |
| `only_addresses_this_transport_can_dial_are_ever_probed` | the owner's LAN / hostname / TLS candidates never dialled |
| `only_an_address_this_transport_can_dial_is_ever_chosen_and_stored` | unit 14's v6 property |
| `the_machine_identifier_is_read_from_json_and_from_xml_alike` | PMS answers XML unless the JSON Accept survives |
| `our_own_server_leads_the_roster_however_plex_tv_ordered_it` | registration order (the registry makes the FIRST registration current) |
| `switching_profile_re_keys_every_source_and_drops_the_ones_not_granted` | per-(user, server) tokens |
| `the_roster_round_trips_through_the_session_file_format` | persistence |
| `a_corrupt_or_absent_roster_never_costs_the_session` | the silent-sign-out class |
| `a_single_server_session_behaves_as_it_always_has` | no behaviour change on one server |
| `a_malformed_home_user_costs_that_tile_and_not_the_session` | same class, existing field |
| `a_hostname_ranks_behind_an_address_that_can_actually_be_dialled` | the probe.rs tiebreak |

## One integration line this PR cannot add

`auth::install_stored_roster(&session)` registers the persisted roster on a resumed session. It is
called from `start_switch`, which covers the boot picker and every "Change profile". The one path it
does **not** cover is `app.rs`'s straight-to-Home boot (a stored session with a single Plex Home
user, or any automated run), which installs the primary itself and never enters `auth`. `app.rs` is
outside this unit's file scope, so it needs one line beside its `install_pms`:

```rust
crate::auth::install_stored_roster(&session);
```

Without it, shares are registered only from the sign-in / picker paths.

**Nothing here is verified on device.** I took an `--fps` run before the fleet-wide lock correction;
it showed the contention signature (6 failures, every one "scene never entered this screen") and is
**void** — not reported as a pass or a failure.

This unit's effect is in the **event log**, not on a frame: discovery runs during sign-in, and the
shared-source UI (Source chip, shelf annotation) belongs to sibling units. So the recipe below reads
the log and the roster, and uses the screen only as a liveness check.

### A. The one that actually exercises discovery (needs a human with a phone, once)

Discovery only runs on the **login** path, so it needs a real QR sign-in.

```sh
ssh root@$TV 'rm -f /tmp/plxnative-events.log; touch /tmp/plxnative-login'   # forces the QR screen
make deploy && make run RUN_SECS=120
# scan the QR with the Plex app and authorize
```

**A correct log carries these lines, in this order** (`<share-ip>` = the share's public IPv4,
`<lan>` = our server's LAN IPv4):

```
auth: authorized — discovering server
auth: resources n=<N> servers=2
auth: reached "<our server>" <lan>:32400 (ours)
auth: reached "<share-name>" <share-ip>:<share-port> (shared by <friend-handle>)
auth: 2 server(s) reached, primary '<our server>'
plex: server slot 0 registered at <lan>:32400
plex: server slot 1 registered at <share-ip>:<share-port>
auth: PMS client installed <lan>:32400
```

**The negatives matter more than the positives.** None of these may appear:

- any address starting `<owner-lan-prefix>` — that is the *owner's* LAN, and dialling it is the 8-second trap;
- an 8+ second gap between `authorized — discovering server` and `2 server(s) reached` (timestamps
  are `SDL_GetTicks` ms; the two dials should total well under 2 s on a healthy LAN);
- `answered as a DIFFERENT machine` — if this appears, an address is being answered by something
  that is not the server, and the verification is doing its job; report the line;
- a v6 literal or a hostname anywhere in a `reached`/`registered` line.

`auth: '<name>' did not answer (N address(es) tried, M not dialable)` is **not** a failure on its
own — it is expected for any server that is genuinely off. It is a failure only if it names a server
you know is up.

### B. What the persisted roster must contain (no sign-in needed once A has run once)

Reads the session file **without printing a token**:

```sh
ssh root@$TV 'grep -o "\"machine_id\":\"[^\"]*\"\|\"address\":\"[^\"]*\"\|\"port\":[0-9]*\|\"owned\":[a-z]*\|\"shared_by\":\"[^\"]*\"" /media/developer/com.beb.plxnative-auth.json'
```

Expected for this account — **two** sources, **ours first**:

| # | `owned` | `shared_by` | address it must have settled on |
|---|---|---|---|
| 0 | `true` | `""` | our server's **LAN IPv4**, port 32400 — *not* the IPv6 literal plex.tv also offers |
| 1 | `false` | `<friend-handle>` | the share's **public IPv4**, port <share-port> — *not* `<owner-lan-ip>`, *not* the internal hostname |

Two **distinct** tokens (check the count, not the values:
`grep -c '"token"' …` should be ≥ 3 — one per source plus the primary `ServerRef`).

### C. Resumed-session boot (proves the roster survives a reboot)

```sh
ssh root@$TV 'rm -f /tmp/plxnative-events.log /tmp/plxnative-login'
make run RUN_SECS=30
```

With a multi-user Plex Home, the boot picker path logs
`auth: roster restored — 2 server(s) registered`. **On a single-user account it will NOT appear** —
that is the missing `app.rs` line above, not a regression.

### D. Liveness (a frame, weak evidence, 20 s)

`tools/capture-screen.sh out.png DISPLAY` after a resumed-session boot. A correct frame is just
**today's Home**: the top tab bar, the hero at the top, poster shelves below with real titles and
artwork. Nothing about shared sources is drawn yet by this unit — if Home populates as it always
did, this unit has not broken the single-server path, which is all a frame can say here.

### DIFF
--- a/rust-modules/src/auth.rs
+++ b/rust-modules/src/auth.rs
@@ -8,8 +8,9 @@
 //! All network happens on spawned threads; the UI only reads snapshots through the accessors here.
 //! Tokens live in the working [`Session`] and are never logged.
 #![allow(dead_code)]
-use crate::plex::account::{AccountClient, HomeUser};
-use crate::plex::session::{self, ServerRef, Session, UserRef};
+use crate::plex::account::{AccountClient, HomeUser, Resource};
+use crate::plex::probe::{self, Candidate, Outcome, ProbePlan, Scheme};
+use crate::plex::session::{self, ServerRef, Session, SourceRef, UserRef};
 use std::sync::Mutex;
 
 /// Which stage the flow is in — the Login/Profiles screens switch on this each frame.
@@ -190,21 +191,35 @@ pub fn submit_pin(index: usize, pin: &str) {
 
 /// Main-loop hook: when the flow has resolved credentials, return them ONCE (and persist the
 /// session) so the caller installs them on the main thread. `None` on every other frame.
+///
+/// The returned creds are the PRIMARY server's, unchanged. The rest of the roster is registered
+/// here too — every path that resolves credentials passes through this one function (sign-in, a
+/// profile pick, and `cancel`'s resume of the stored session), and a share that is not in the
+/// registry is a share nothing can browse.
 pub fn take_ready() -> Option<ReadyCreds> {
-    with_ctl(|c| {
+    let (sources, creds) = with_ctl(|c| {
         if c.phase == Phase::Ready && c.apply_pending {
             c.apply_pending = false;
             session::save(&c.session);
             session::set_current(Some(c.session.user.clone())); // drives the Home profile chip
-            Some(ReadyCreds {
-                host: c.session.server.address.clone(),
-                port: c.session.server.port as i32,
-                token: c.session.pms_token().to_owned(),
-            })
+            Some((
+                c.session.sources.clone(),
+                ReadyCreds {
+                    host: c.session.server.address.clone(),
+                    port: c.session.server.port as i32,
+                    token: c.session.pms_token().to_owned(),
+                },
+            ))
         } else {
             None
         }
-    })
+    })?;
+    // Outside the CTL lock: registering touches the server registry (and, on a cold slot, reads
+    // the session file for the device id), and nothing here needs the flow state held while it
+    // does. `None` for the primary — the caller's own `plex::install` of these creds is what
+    // retargets `current`, and an owned entry registers first regardless.
+    install_roster(&sources, None);
+    Some(creds)
 }
 
 /// Open the "who's watching" picker: the boot gate (picker-at-start) and the Home profile menu's
@@ -217,6 +232,11 @@ pub fn start_switch() {
     if sess.account_token.is_empty() {
         return set_error("You're signed out — sign in to use profiles.");
     }
+    // The boot picker reaches this before any profile is chosen, so it is the earliest point on
+    // the resumed-session path where the stored roster can go back into the registry. Idempotent,
+    // and it does not touch `current` — a "Change profile" from Home lands here too, by which time
+    // everything is registered already and this is a no-op.
+    install_stored_roster(&sess);
     with_ctl(|c| {
         c.error.clear();
         if c.users.is_empty() {
@@ -315,8 +335,22 @@ fn login_thread() {
         return log("auth: sign-in was cancelled while the pin poll was in flight — token dropped");
     }
     let ac = AccountClient::new(&cid, Some(&token));
-    if !discover_and_store(&ac) {
-        return set_error("No local Plex server found on this network.");
+    // The failure copy is per outcome, and it used to be one line — "No local Plex server found on
+    // this network." — for every one of them. That sentence was the discovery POLICY talking: a
+    // server reached over the internet was a failure by construction, so the message named the LAN.
+    // It now describes what actually happened, and none of the three sends the user to the wrong
+    // place: a token refusal is not a router problem, and an account with no server is not an
+    // outage.
+    match discover_and_store(&ac) {
+        Discovery::Ok => {}
+        Discovery::NoServers => return set_error("This Plex account has no server yet."),
+        // "A Plex server", not "your": `refused` is raised by ANY server's addresses all answering
+        // 401, which includes a friend's share while our own was merely silent. Naming the wrong
+        // machine is the whole failure mode this per-outcome copy exists to avoid.
+        Discovery::Refused => return set_error("A Plex server refused the connection — check its network access settings."),
+        Discovery::Silent => return set_error("Couldn't reach any Plex server — check the connection."),
+        // Silent on purpose: the user pressed BACK and is already on the screen `cancel` resumed.
+        Discovery::Cancelled => return log("auth: sign-in was cancelled while discovery was dialling"),
     }
     // Install the PMS client now (server/owner token) so the who's-watching avatars can proxy
     // through the server's photo transcoder. The per-user token is swapped in on profile pick.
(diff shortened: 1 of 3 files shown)


GLinnik21:

PR #30, opened 2026-08-13
ui: the first-run question — what goes on your Home, asked once

GLinnik21:
Deliverable **F** of the Shared Sources design: one question, asked once, after the profile picker
and before Home — *"What goes on your Home?"* — and **only when the roster holds more than one
source**. A single-server install (ninety per cent of them) goes straight to Home exactly as it does
today.

Nothing here grants or blocks access. Every library it lists stays browsable from the Library chip
either way; it decides one thing, what reaches **Home**.

## What it is

- **A route, not a sheet over content** — the standing sign-in and the picker have, which is what the
  no-full-screen-sheets rule protects. Two columns on the app's own ground: copy + the single action
  left, the rows right, **no panel, no radius, no shadow** (a panel's depth claims a screen behind it;
  a route has none).
- **Geometry from the design system's route frame**, added to `ui/consts.rs` as documented constants
  (`ROUTE_TOP`, `ROUTE_COPY_X`, `ROUTE_COPY_W`, `ROUTE_ACTION_BOTTOM`, `ROUTE_LIST_X`,
  `ROUTE_LIST_W/H_MAX`). **One** top guide at 150 for both columns: the canvas writes two separate
  `--route-copy-top` / `--route-list-top` vars the design system never defines, and its own
  route-screen card uses the single `--route-top` for both.
- **Own libraries arrive `On`, a friend's `Off`** — the word at the trailing edge of the same
  `TableView` toggle rows the Sources list's *On Home* level uses. Group header = the **machine**,
  accessory = the **person**; the copy names the people.
- **Focus starts on *Start watching*** (fast path = one press; the rows are one press right).
  **BACK skips**, and records exactly what is on screen — for an untouched screen, that default.
- An **unreachable** source keeps its place and its name, dimmed as a whole group at
  `theme::UNREACHABLE_A` (0.52), accessory `"<person> · Can't be reached"`.

## The state the rest of the feature reads

`first_run::is_pinned(machine_id, key, own)` — the one place a recorded decision and the arrival
default meet. An entry when somebody decided; the ownership default when nobody has. That single rule
is what makes a share arriving *later* turn up off Home, and a new library of your own turn up on it.
Persisted as `session::PinRef { machine_id, key, on }`, keyed by **machineIdentifier + section key**
(both servers in the live capture call their first library `1`). A server that will not answer
`/identity` records nothing and is drawn as unreachable — a pin keyed `""` would be a pin on that
section number of every unnamed server at once.

Its roster comes from `app.rs::note_sources` (current server + every `/tmp/plxnative-servers` entry).
**That function is the seam the plex.tv roster replaces** — when the roster unit lands it calls
`first_run::add_source` and nothing about the screen changes.

## Shared components improved rather than forked (`ui/table.rs`)

| | |
|---|---|
| `Section::accessory` | declared and never drawn until now — the header's right-hand run |
| `Section::dim` | a whole group at one alpha, **header included** (`Row::dim` is an ink swap, and a header is already tertiary grey) |
| `TableView::focused` | is the ring in this list? `false` → the faint selected plate instead of the accent fill, so a list beside a control does not draw a second ring |
| `TableView::empty_note` | `None` = "my headers are my content while the servers answer" — the old placeholder covered them *and* measured the frame at 120px, scissoring two groups of three off the bottom |

Also `fmt::library_count` ("412 films" / "238 shows"; a negative count is *unknown*, and draws nothing).

## Host gate

`make check` **three consecutive green runs, 432 passed** (418 before: 11 in `ui::first_run`, 1 in
`ui::fmt`, 2 in `ui::table`). `make` (ARM) links; `--no-default-features` (the RELEASE feature set)
compiles. `code-review` run at high effort — every finding in my own files fixed (see below).

Tests pin the things that are policy rather than code: the gate (**one source → never; two → once,
and not again on the next boot**, driven through a real `restore()` from a saved session), the
defaults, that BACK's honest default is what gets recorded, that a flip moves exactly one row, that
an unidentified server records nothing, and that focus starts on the action with the rows one press
right.

**Not verified on device.** No frame of this screen has been photographed; the fleet's single
television is held for a merged-stack pass. Everything below is the recipe to do it without me.

**Why it needs a recipe at all:** the screen sits immediately downstream of the who's-watching
picker, and *any* `/tmp/plxnative-*` trigger (bar the documented DIAG exemptions) marks the boot
automated — which suppresses the picker **and**, by the same rule, this question. So it must be
forced.

**Arm (inside the lock, after clearing `/tmp/plxnative-*`):**

| file | content |
|---|---|
| `/tmp/plxnative-token` | the owner PMS token (what `tests/run.py` injects) |
| `/tmp/plxnative-servers` | `[{"name":"bx23-ldn","owner":"bamx23","machine_id":"<share-machine-id-redacted>","host":"<share-ip-redacted>","port":<share-port-redacted>,"token":"<the share's accessToken>"}]` |
| `/tmp/plxnative-firstrun` | empty — **forces the question** whatever the session says |

The second server's `accessToken` is resolvable exactly as the harness does it — `run.py`'s
`resolve_shared_server()` against plex.tv `/resources` with the owner token from
`src/config.local.h`; the block it needs in `tests/manifest.local.json` is
`"shared_server": {"machine_id": "<share-machine-id-redacted>", "name": "bx23-ldn"}`
(this account's real share, `sourceTitle` = `bamx23`; plex.tv offers it no LAN address, so
`<share-ip-redacted>:<share-port-redacted>` — plain HTTP on a numeric IP, which is all `stream.rs` can dial — is the one
that answers). **Without `plxnative-servers` there is only one source and the screen has nothing to
ask about**, which is itself the single-server regression check: with only the token armed, the app
must land on Home.

To see it in the *interactive* position (after the picker) rather than forced: arm
`/tmp/plxnative-pickuser=<index>` alongside the two credential files — that forces the picker on an
automated boot and auto-picks a tile — and the question must follow the pick, before Home. That path
also exercises the login-landing gate (`landing_route` at `take_ready`), which is the one the
picker's users actually take.

**Drive it** through `/tmp/plxnative-remote`: `right` (ring leaves the action, enters the rows),
`down`, `ok` (the focused row's word flips `On`↔`Off`), `left` (back to the action), `ok` (answers,
routes to Home). `ck:X,Y` clicks work too, in authored 1920×1080 coords.

**Capture** with `tools/capture-screen.sh out.png DISPLAY` and check:

1. **As it first appears.** Heading *"What goes on your Home?"* at HERO, cap-top on **y=150**, left
   edge **x=90**, wrapping inside 760px. Under it, one paragraph naming the PEOPLE — *"bamx23 has
   shared libraries with you. Pick the ones you want on your Home screen — you can browse any of
   them from the Library chip whenever you like."* — no machine name anywhere in it. The pill
   *Start watching* is **focus-filled** (accent, near-black ink), left edge x=90, its **bottom edge
   at y=1006**. The rows start at **x=930**, right edge at x=1830, top **y=150**.
2. **The rows themselves.** Group header = the machine (`Gleb's Mac mini` — or the bare address on a
   token boot — and `bx23-ldn`), accessory right-aligned on the same line (`This account`,
   `bamx23`). One row per library, sub-line `412 films` / `238 shows`, and the word **`On` for your
   own, `Off` for the friend's** at the trailing edge. **No sheet under any of it** — the ground
   behind the rows must be the same flat `#2C2C2E` as behind the copy, with no rounded corner and no
   shadow edge anywhere along the column. The selected row wears the **faint** plate, not the accent
   fill, while focus is on the action.
3. **Focus moved right.** After `right`, the first row's pill becomes the **accent** fill with dark
   ink and the *Start watching* pill drops to its idle face. Exactly one bright fill on screen at
   any moment.
4. **A row toggled.** After `ok` on the friend's row, its trailing word reads `Off`→`On` and nothing
   else on screen changes.
5. **Answering.** `left`, `ok` → Home, and the event log carries
   `firstrun: answered — <n> of <m> libraries on Home`. Relaunch with the same triggers **minus**
   `plxnative-firstrun`: it must go straight to Home (the question is answered — `home_pins_asked`
   in the session file).
6. **Unreachable.** Point the `plxnative-servers` entry at a dead address (`"host":"10.255.255.1"`).
   The group must still be listed, header and person and all, at 52% with `· Can't be reached`
   after the person, with no rows under it — and the rest of the screen untouched.

Log lines to grade against: `servers: #0 name=… owner=… … creds=ok`, `firstrun: asking about N
sources`, `firstrun: own server libraries=<n>` / `shared server libraries=<n>`, `route=firstrun` on
any FRAMEDROP line, and the `firstrun: answered` line above.

`./tests/run.py --fps` was **not** run (no lock, and the tier has no scene for this route). The
screen is static once settled, so if a scene is wanted later it wants an `fps_ceiling`, not a floor —
the only animator on it is the loading spinner, which reports to `ui::idle` by being drawn.

## Note for the coordinator — findings outside my files

`code-review` surfaced three in the base branch that I did not touch, the first of which is HIGH:

- **`plex/account.rs:106`** — `includeIPv6=1` now asks plex.tv for v6 connections, but the live
  consumer (`auth.rs:392`) stores the first `local && !relay` connection with no address-family test,
  and `stream.rs` is `AF_INET`-only. A stored `2001:db8::…` passes `can_go_local()`, so every later
  boot reports "local server" and nothing ever loads. `probe.rs`, which ranks v6 last, still has no
  callers.
- **`plex/models.rs:596`** — `de_bool` has no fallback arm (`de_i64`/`de_str` both have one), so a
  float/object/array encoding still fails the whole `/resources` array, which is the blast radius
  that adapter exists to prevent.
- **`tests/run.py:1519`** — with no `shared_server` overlay the FPS path reports "no FPS scenes
  defined in the manifest" (pointing at the manifest, not the missing block) and drops the skipped
  scenes from the summary; the case path has a dedicated message for exactly this.


GLinnik21:

PR #31, opened 2026-08-13
library: the section table learns about servers, and the toolbar gets its Source chip

GLinnik21:
Deliverable **A** of the Shared Sources design: the Library's section table goes multi-server, and the toolbar gets its Source chip and the two-level Sources list. Deliverable **B** is answered by subtraction here too — the tab strip keeps naming types, not people.

## What changed

**The table is addressed by (SOURCE, section).** Both servers have a section `1` and they are different libraries, so a bare key names two things. Every fetch is issued through `client_for(sid)` **captured at the spawn site** — the rule `browse.rs` already stated for its page worker and did not keep.

**It grows by APPEND, never by rebuild.** A page landing is blamed on a section INDEX, so a table that reshuffled under an in-flight fetch would splice one library's items into another's store. `ensure_sections`'s early-return was the only thing preventing that; appending replaces it and holds for every source. Two generations now: `SECTIONS_GEN` (shape, moves on an append — what the label caches key on) and `EPOCH` (identity, moves on `reset` alone — what index-blamed landings gate on), so one source's append cannot discard another's answer.

**Only the current server is discovered on the main thread.** `ensure_sections` runs at boot and on every Library entry; fanning out there would park the SDL loop for one `connect(2)` per unreachable share. Every other source is discovered by a worker off `pump` — sections, then the server's own `friendlyName`, then a `size=0` count probe per library — with a 10 s per-source backoff.

**The Source chip**, first in the toolbar: `Library · LDN Films  bamx23 ▾`, the handle a `MICRO` run at 62% of the label's own ink, **absent entirely on your own libraries**. With one source the chip is not drawn and the list does not exist. Its 640-wide panel has two levels swapped by pills at the top: *Browse* (one tick, no words, OK closes) and *On Home* (the word `On`/`Off`, OK flips, the panel stays open). Groups are servers — machine as header, person as accessory. Last pinned dims its value and states the rule. An unreachable group dims whole and still reads `On`.

**The toolbar is a typed chip list.** `on_ok` matched on the chip's INDEX with a `_` catch-all, so inserting Source at 0 would have made it open Sort while Sort and Filter both opened Filter — silently. `open_chip_menu` is the one exhaustive chip → action map and the pointer path calls it directly.

**Picking a library on another server moves the app's current server** (§5's named *cheap variant*). Without it the panel is a trap: `PmsMovie` carries no `ServerId`, posters and ratingKey resolution go through `client()`, and ratingKeys are server-local — so OK on a friend's card would open and play a different title of yours with the same number. Home's catalog is dropped and re-armed asynchronously; `route`'s cached `machineIdentifier` is forgotten.

**`TableView`** gained the two things it lacked: `Section::accessory` was declared and **never painted**, and `Section::dim` is one alpha over header, accessory, rows and the selection pill.

## Host gate

- `make check` **green three times**, 419 → **431** tests.
- `make` and `make RELEASE=1` both link (ARM).
- `code-review high` run; all seven findings addressed (four were latching bugs of the same shape as this module's known single-flight one — an empty count landing latching the probe off, `reachable` unable to flip back after discovery, "Check for new shares" only re-asking known-dead sources, and the chip mixing a queued name with a committed handle). The one finding left open is pre-existing and belongs to the probe/ingest unit: `auth.rs:396` has no address-family filter, so `includeIPv6=1` can now store an address `stream.rs` (AF_INET, dotted-quad) can never dial.

**Not verified on device.** The television was unavailable for the whole of this work; nothing below has been seen.

### Setup

The second server must be injected, since nothing else registers one yet.

1. Resolve the share's credentials from plex.tv with the owner token (never printed):
   ```
   python3 - <<'EOF'
   import json,os,sys
   REPO="<worktree>"; sys.path.insert(0, os.path.join(REPO,"tests")); os.chdir(REPO)
   import run as h
   t=h.read_token(); r=[x for x in h._server_resources(h._plextv_resources(t)) if not x.get("owned")][0]
   host,port,_=h.pick_connection(r)
   print(json.dumps([{"name":r["name"],"machine_id":r["clientIdentifier"],
                      "handle":r.get("sourceTitle") or "","host":host,"port":int(port),
                      "token":r["accessToken"]}],separators=(",",":")))
   EOF
   ```
   For this account that resolves to **`bx23-ldn`, handle `bamx23`, `<share-ip-redacted>:<share-port-redacted>`** (plain HTTP; `httpsRequired=false`).
2. `ssh root@$TV 'cat > /tmp/plxnative-servers'` with that JSON, then `tools/tv-session.sh up --screen library --keep` (`--keep` so the clear does not remove it). The app registers it in `install_pms` and logs `servers: #0 name="bx23-ldn" handle="bamx23" …` and then `browse: roster now 2 source(s)`.

Give the friend's libraries a few seconds to arrive: their discovery is a WAN worker driven by `pump`, which only runs while the Library screen is up.

### The six frames, and what a correct one looks like

`tools/capture-screen.sh out.png DISPLAY`, then look at it.

| # | How to get there | Correct frame |
|---|---|---|
| 1 | `up --screen library` with **no** `plxnative-servers` | Toolbar at y≈134 has **exactly two** chips from x=90: `Sort · Title`, `Filter · All`. **No Source chip, no empty slot.** |
| 2 | with the share armed, boot to library (your own Movies) | **Three** chips; the first reads `Library · <your library>` with a chevron and **no handle run at all** — this is the one to look at hardest, because an empty run would show as extra width before the chevron. |
| 3 | `key up` (grid → toolbar, focus lands on Source), `key ok` | 640-wide panel at x=90, y=204. Two pills at its top: **Browse** filled (subtle white pill, primary ink), **On Home** bare tertiary text. Below: one group per server — header `Gleb's Mac mini` with **no** accessory, header `bx23-ldn` with `bamx23` right-aligned on the same line one rung down. Rows are two-line (`Movies` over `185 films`). **Exactly one tick**, on the library you are browsing. **No `On`/`Off` anywhere.** Last row, under a hairline: `Check for new shares`. |
| 4 | `key up` (onto the pills), `key right` | Same panel, same height, **nothing has moved**. **On Home** now filled, **Browse** bare. Every library row reads `On` or `Off` at its trailing edge — your own `On`, the friend's `Off`. **No ticks anywhere.** |
| 5 | `key down` (into the rows), `key ok` on your first library, then `key down` | The row you unpinned reads `Off`. The remaining pinned row's `On` is **dimmer than the other rows' words** while its label stays full-strength, and its sub-line reads `Home needs one library` instead of its size. `key ok` on it must do **nothing at all** — no flash, no toast, the word does not change. |
| 6 | with the panel open, make the share unreachable — on the TV: `ip route add blackhole <share-ip-redacted>` — then `key ok` on `Check for new shares` and wait ~10 s | The `bx23-ldn` group **dims whole at .52 — header included**, its accessory reads `Not reachable · bamx23`, and its rows **still read `On`**. Your own group is untouched. (Undo with `ip route del blackhole <share-ip-redacted>`.) |

Also worth one look, since it is the riskiest interaction: on the **Browse** level, `ok` on a row of the friend's group must close the panel, dissolve the grid, relabel the chip to `Library · <their library>  bamx23`, and leave the **type** pill selected in the top strip (their films keep *Movies* lit). The event log should show `browse: current server is now source 1`.

### FPS gate

`./tests/run.py --fps` once, at the end. The library scenes arm `/tmp/plxnative-library` — note `=N` now names a tab **pill**; with one source that is unchanged.


GLinnik21:

PR #32, opened 2026-08-13
identity: an item is a (server, key) pair, not a bare ratingKey

GLinnik21:
Plex numbers every item-shaped id from 1, **per server**. `ratingKey`, `librarySectionID`,
`Part.key`, `Stream.id`, `personId` — all server-local integers. With one server that costs
nothing and the app has always leaned on it. With a share registered it is not an edge case
but the **measured default**: `docs/shared-servers.md` §2 records the friend's server
answering with section key `1`, exactly like our own `Movies`, and its ratingKeys starting
at 1 alongside ours.

This makes an item a `(ServerId, key)` PAIR at the five places that identified one by a
bare key, each of which is otherwise a confident answer about the wrong machine.

## The five sites

| site | what a bare key did |
|---|---|
| `pms::index_of_rk` | linear `m.rk == rk` over one flat catalog → returns whichever row is EARLIER. Callers: the item menu's **Play from Start** (long-press a friend's episode on a merged Continue Watching shelf → *our* film with that number plays, friend's title still in the HUD) and `detail::mount_rk` (wrong backdrop, blur envelope, selection). |
| `metadata::cached_playing` | **the worst, because it is silent and it feeds playback** — the fast path that SKIPS the PMS fetch. A false hit substitutes the loaded page's whole `PlayingItem`: its audio/subtitle `Stream.id`s (then PUT to the *other* server), its frame size (the direct-play gate reasons about the wrong resolution), its fps, chapters and markers. |
| `metadata::pump_season`'s `d.rk != r.rk` | hop A's show → B's show with the same rk while a `/children` is in flight, and A's episodes install onto B's page. Nothing else can see it: the hop bumps no generation that separates them. |
| `plex::queue_index_of`'s rk FALLBACK | the primary key `playQueueItemID` is minted by the queue and is safe; the fallback compares a key the caller supplied, which can come from anywhere. |
| `trail::Node::Detail{rk}` / `Node::Person{key}` | the trail stores identity precisely because indices go stale — and then stored one that had gone ambiguous. Browse the share, open an item, go deeper, press BACK, land on a page the user has never seen. |

## How

**One rule, one place.** `plex::same_item((sid, key), (sid, key))` — so the five sites cannot
drift, and one host test grades the rule itself. `ServerId::UNSET` matches only itself: that
is the pre-registry state (and every row a host test builds by literal), **not** a wildcard,
so the single-server app is byte-identical and a "server unknown" row never answers for a
named machine. Every miss degrades safely where it lands — a catalog miss opens the page
off-catalog, a playing-item miss costs one round trip, a season miss leaves the page it
found alone.

**Stored rows gain the server they came from**, stamped where they are parsed:
`PmsMovie`, `Detail`, `PlayingItem`, `QueueRow`, `person::Person`, `trail::Node`.

**The server is PASSED IN, never looked up.** `parse_item` runs on the hub, page and person
workers and `fetch_full` on the detail worker; the house rule (browse.rs's spawn site states
it outright) is that a worker reads no statics. It is also the only correct answer — the
current server can move while a fetch is in flight, and the rows in hand belong to the
machine that was asked. So the fetch chain takes a `ServerId` and dials `client_for(sid)`
instead of `client()`, which is what makes a row's stamp provable rather than incidental.

**A person prefers the `tagKey`.** That guid is plex.tv's and therefore GLOBAL — one guid is
one person however each machine numbers them — so `Node::same_page` / `trail::same_person`
match on it whenever both sides have one, and fall back to the server-scoped `(sid, key)`
only when a credit row carried none. A present guid is deliberately never matched against a
missing one, or every guid-less row would be the same person as every other.

**`Trail::reset` was not the answer for a server switch.** Its doc calls a truncation the
profile-switch guard ("a new identity must never be able to walk back into the previous
one's pages") and a source switch has the same shape — but not the same remedy, because a
source change does not revoke access to what is behind you and the trail's other rule is
that a cycle is a HISTORY. A per-node `sid` keeps the history walkable and still lands every
pop on the page it names.

**Two dropped writes now say so.** `client_for(sid)` can return `None` where `client()` used
to panic, and the item menu repainted either way — so a scrobble or a deck removal that never
reached a server is a log line rather than the one PMS write with no trace at all.

## Deliberately NOT in this change

`route.rs` still resolves the **stream** against the current server (`ResolveEnv::snapshot`
takes `plex::current_server()`), so `/decision`, the PlayQueue `server://` uri and the stream
host are unchanged. That is the sibling unit's seam — this unit owns the equality LOGIC
inside `cached_playing`/`fetch_playing_item`, not the play path's routing — and the comment
at the site says so rather than implying the play path is finished. Until that unit lands,
a play from a share resolves against the wrong host; the *row lookup* and the *track store*
are what this change makes honest.

`browse::BrowseSection.key` (the verified section-`1` collision) is likewise its own unit;
the rows it produces are stamped here, the section table is not.

## Tests — 427 host tests, `make check` green ×3

Every one of the five sites has a collision test that constructs the SAME rk on DIFFERENT
servers and asserts it never aliases:

- `plex::servers` — `one_rating_key_on_two_servers_is_two_different_items` (the rule, incl. the UNSET semantics on both sides)
- `pms` — `a_catalog_row_is_found_by_its_server_and_key_never_by_the_key_alone` (ours first in the catalog, so a bare-key scan would always answer with it)
- `pms` — `every_row_a_hub_fetch_builds_is_stamped_with_the_server_it_was_asked_of` — **the linchpin**: without the stamp every row is UNSET, every pair compares equal, and each test below passes while production aliases. Graded through the shipped `build_hubs` from a real `/hubs` body.
- `metadata` — `the_playing_item_cache_hits_only_for_the_same_item_on_the_same_server`
- `metadata` — `a_season_landing_for_another_servers_show_with_the_same_key_is_refused` (with the control: the same landing DOES install when the page is still ours)
- `plex::timeline` — `the_rating_key_fallback_is_scoped_to_the_queues_own_server` (and the `playQueueItemID` primary still resolves across servers, as it must)
- `ui::trail` — `one_rating_key_on_two_servers_is_two_pages_in_the_history`, `a_person_is_the_same_page_by_guid_and_a_local_id_is_server_scoped`, `same_page_ignores_the_place_and_never_crosses_kinds`

`metadata.rs`'s tests hold `lib.rs`'s crate-wide `testlock::serial()` (the detail and season
mailboxes contend across modules); `pms.rs`'s hold it too, since `ui::home` walks the catalog
statics. The `trail`/`timeline`/`servers` identity tests touch no global and stay parallel —
`ServerId::from_raw` is now `const` so a fixed slot can be a `const` in a test.

`make` (the ARM cross-link) succeeds.

I ran `./tests/run.py --fps` twice while holding `/tmp/plxnative-tv.lock`; the second run was
**11 passed / 0 failed of 11** (the first showed one `home-grid` "scene never entered this
screen", which the re-run did not reproduce). Per the fleet-wide correction those results are
**not** offered as evidence — treat this branch as **not verified on device** and re-take it
on the merged stack.

**What to run, and why each one:**

1. **`./tests/run.py --fps`** — covers every surface this change threads a `ServerId` through:
   `detail-transition`, `item-menu`, `person-page`, `home-detail-nav`, `home-library-nav`.
   The nav scenes matter most: they bounce the ROUTE through the real press path, so they
   exercise `to_detail` → `Node::Detail{sid, rk}` → `enter_node`'s re-open guard → `trail.back()`.

2. **Two playback cases** — `cached_playing` is the one thing here that reaches the Load
   payload, so prove the pairing did not break it end to end:
   `./tests/run.py --filter dp_h264_ac3_1080p` (direct play) and
   `./tests/run.py --filter transcode_av1_no_dp_audio` (transcode).
   *(Note `--filter` did not narrow the FPS tier for me — it ran all 11 scenes. Check it
   narrows the player tier before trusting a short run.)*

3. **A detail page, and BACK out of it.**
   Arm `/tmp/plxnative-detail=<any ratingKey in the home catalog>` and capture with
   `tools/capture-screen.sh out.png DISPLAY`.

**What a CORRECT frame looks like** — the failure this change prevents is *the wrong item
drawn confidently*, so grade CONSISTENCY, not prettiness:

- The detail page's **hero backdrop, poster, blur/ambient wash and title must all describe
  the SAME film**. A regression here is a page whose title says one thing and whose backdrop
  and colour wash say another (`mount_rk` resolving the catalog row by a bare key).
- The **item menu** (`/tmp/plxnative-itemmenu`, `route=itemmenu`) must name the card it is
  anchored beside — the rows adapt per item kind, so an episode card must show
  `Go to Episode / Go to Show`, a movie card `Go to Movie`.
- **BACK from a detail page must return to the page it was opened FROM**, not to Home. With
  `/tmp/plxnative-navosc=<ratingKey>` the route should bounce Home ↔ that detail page every
  1400 ms forever; a stall on either side, or a landing on a third screen, is this diff.
- The **person page** (`/tmp/plxnative-detail=<rk>`, `plxnative-detailsec=1`,
  `plxnative-detailok`) must show the actor whose headshot was focused, with their own
  Movies/Shows shelves — not an empty page and not someone else.
- In the event log: `detail: rk=… '<title>' …` should name the rk that was armed, and
  `playing item: rk=… audio=N subs=N markers=N chapters=N` should be non-zero for a case
  whose file has tracks. `audio=0` on a file with audio is `cached_playing` missing where it
  used to hit — which is *safe* (it falls through to the fetch) but worth noticing.

Only one physical TV exists and there is no host runtime, so none of the above is reachable
from `make check`.


GLinnik21:

PR #33, opened 2026-08-13
pms: Home is a merge of every source, and one dead share cannot take it down

GLinnik21:
> **Merged `shared-servers-integration` (twelve units + device fixes) into this branch.**
> This branch now CONTAINS integration, so the diff against `shared-servers-base` is everyone's
> work. **The reviewable diff for this unit is `shared-servers-integration...HEAD` — three files:
> `pms.rs`, `browse.rs` (one doc comment), `ui/home.rs` (two constants).** Integration is not on the
> remote, so the PR base cannot point at it.
>
> **526 host tests green (3/3 runs), ARM links.** Not device-verified — the coordinator has the TV.

---

> ### ⚠️ Read before device-testing the merge
>
> **A share contributes to Home only once one of its libraries is PINNED, and a friend's libraries
> start unpinned** (`browse.rs:543` — "your own libraries feed Home; a friend's do not until you say
> so", the design's first-run state). That is deliberate and it is the official client's behaviour:
> canvas C says "the merge is conditional on pinning — confirmed in the real app: a borrowed film
> joins Continue Watching only once its library is pinned."
>
> **The UI that does the pinning is unit 6 (#36), which is not merged yet.** So on a device test
> today the friend's shelves will be absent from Home and it will look exactly as if this unit had
> not landed. It has: `feeds_home` is reading a pin store whose only writer is still unreachable
> from the remote.
>
> Two ways to see the merge before #36:
> - flip the default at `browse.rs:543` to `true` for one run (a share starts pinned), or
> - call `browse::toggle_pin(i)` on the friend's section from any debug path.
>
> Whether to ship the first-run default as unpinned (design) or pinned (what the owner asked to
> *see*) is a product call, and it is one line in `browse.rs`, not in this unit.

---

## Pinning — which libraries reach Home

The data layer for the one user-facing control shared servers introduce. A per-library `pinned`
flag, persisted in the session, with **Home as its only reader**.

Three states, kept apart deliberately:

| state | who decides | what it governs |
|---|---|---|
| **granted** | plex.tv | everything browsable. Not a setting — the app has no server on/off at all |
| **pinned** | the user, in the Sources list | Home, and nothing else |
| **reachable** | right now | nothing; an unreachable pinned server still reads `On` |

### What reads the pin

`pms::build_hubs`, once, and nothing else — asserted two ways. Behaviourally: one
`/library/sections` answer and one hub answer, and the unpinned library is gone from Home while
still standing in the directory the section table is built from. Structurally: a test reads
`browse.rs` — which owns the section table, the grid's paging, the server-driven sort menu and the
A–Z rail — and fails if it so much as names the pin model. `ui/library.rs` is deliberately *not*
probed: the Sources list lives in that toolbar by design and is the one screen that must ask.

The gate is per **item**, not per shelf, because the merged Continue Watching row draws from every
source at once: a borrowed film joins it only once its library is pinned (design §C). The hero pool
is built from the shelves below it, so it inherits the same gate for free.

### The rules, and where each lives

- **Defaults** (`pin::resolve`) — your own libraries pinned, a friend's not.
- **The last pinned library cannot be unpinned** (`pin::apply`) — in the data layer, so no UI can
  route around it. The list still draws the row live; only its *value* dims.
- **Empty means unknown, twice.** An empty persisted table = nobody has answered → defaults apply.
  An unresolved roster = we have not been told what exists → the gate passes everything. Either
  read the other way is a blank Home on the first boot after an upgrade.
- **A corrupt table degrades per row**, never failing the `Session` parse — that failure is a silent
  sign-out (`peek` falls through to a default, `load` mints a fresh client id over the real one).
- **The rescue.** The user cannot reach an empty pin set, but the world can: an account with no
  libraries of its own, or the grant behind your only pin disappearing. `resolve` un-empties it,
  reaching for your own library before a stranger's, and taking back exactly **one** answer when it
  has to overrule one.

### Two things this needed from its neighbours

- `Metadata::librarySectionID` (+6 lines in `models.rs`) — half of a library's key. Verified present
  on every hub row and every Continue Watching row of a live PMS, as an int, with the string form
  the OpenAPI spec also documents covered by `de_i64`.
- `session::set_pins` — **one writer** for `Session::pins`. Every other writer saves the whole
  struct from a snapshot of unknown age (`auth` takes one when a profile switch starts and writes it
  back when the roster lands), so a merge rule keyed on "empty means it doesn't know" only covers
  snapshots older than the *first* answer. An owner covers every snapshot of any age.

### Integration seam for the sibling units

- The roster unit calls `pms::pin::install_roster(&[Granted { key: LibKey { machine_id, section },
  owned }])` on every roster landing. `machine_id` is `Resource::client_identifier` (= the registry
  key, = `Client::machine_id`), `owned` is `Resource::owned` — plex.tv's own word, not a guess off
  `sourceTitle`. It writes nothing: resolving is not answering.
- The Sources list walks `pin::snapshot()` (owned data — there is deliberately no borrowing
  accessor, see the comment where one would go) and presses `pin::set(&key, on) -> Set`.
  `Set::LastPinned` is the refusal the `valueDim` treatment renders.
- The first-run route gates on `pin::asked()` and calls `pin::acknowledge()` on **both** exits —
  `Start watching` and the skip. It returns `false` if nothing was written.
- Home needs no wiring: `pin::set` bumps a generation and `pms::pump` refetches, so several toggles
  in one visit coalesce into one fetch and no keypress blocks the loop.

### Tests

437 total (+19 here), all pure — no session file, no socket, no TV:

- the defaults on a fresh roster; a recorded answer beating the default, keyed per server
- the last-pinned refusal, and that pinning a second library frees the first
- the rescue: a borrowed-only account, own-before-stranger, and one answer taken back rather than every answer
- unknown-is-not-none for both an unresolved roster and an unnamed library
- Continue Watching merging only pinned libraries; an unpinned library's shelf absent from Home, hero included
- `home_consults_the_pin_and_nothing_else_does` — the negative rule, behavioural + structural
- a revoked share's persisted row dropped rather than panicking
- the write/read round trip through the session's own serde, including that an unkeyed row is never written
- a corrupt table (six shapes) degrading to defaults with the client id and account token intact
- one broken row costing only that row; no other writer able to revert the table

## What the merge changed in this unit

Three architectures met in `pms.rs`; all three survive.

- **Unit 3's identity stamping** (`PmsMovie.sid`, `plex::same_item`) is kept whole — and it is what
  lets the merged catalog hold two different films both numbered `1` without either aliasing the
  other. `project` takes the `sid` the spawning frame captured, exactly as `build_hubs` did.
- **Unit 12's `HeroSlot` + `own_items_first`** are kept whole, with one correction the merge forces:
  a slot's handle now comes from the **row's** source, not from its shelf's. The merged deck's own
  `source` is empty by design, so a slot reading its shelf would attribute every borrowed film in
  Continue Watching to nobody — and `own_items_first` would conclude one of ours had already opened
  the door. Unit 12's four pool tests are unchanged and still pass.
- **This unit's per-source fetch, latch, ladder and verdict** are unchanged in shape.

**`Source`/`set_sources` are gone, and so is the `app.rs` registration this unit had added.** The
registry is the granted roster and now carries the handle (`ServerFacts`), and integration's
`install_pms` already registers and describes every injected server — doing it twice would have been
two models of one fact. The roster is now derived from `plex::server_ids()` + `server_facts()`.

**`feeds_home` is the seam the pin store fills** (unit 6 / #36), reading `browse::pinned_libraries`.
Its one subtlety is written as a test: an **empty** pinned set means "no section has been discovered
yet", never "nothing is pinned" — `/library/sections` lands after `pms_fetch_hubs`, and
`is_last_pinned` forbids emptying the set. Read the other way, Home would have no sources at all on
the frame it boots. `pinned_libraries`' `#[allow(dead_code)]` is removed: it has a caller now.

**Not verified on device.** No frame of this was seen on the TV: I never held the lock, and per the
coordinator's instruction I did not drive the television. Everything above is host-side.

The unit is inert until a roster is installed, so there is a dev trigger for exactly this. All
values below are placeholders — substitute from `tests/manifest.local.json` / `src/config.local.h`.

**Recipe** (one build, one unbroken locked window — the pair is only evidence if nothing redeploys
between the two captures):

```sh
make deploy                                   # dev build; RELEASE=1 compiles the trigger out
ssh root@<tv-host> 'rm -f /tmp/plxnative-*'   # a stale trigger silently changes the boot screen
ssh root@<tv-host> 'printf %s "<pms-token>" > /tmp/plxnative-token'   # from src/config.local.h

# A — control: the roster is installed, nothing is excluded
ssh root@<tv-host> 'touch /tmp/plxnative-shared'
make run && tools/capture-screen.sh a-control.png DISPLAY

# B — one library treated as a friend's share, so it arrives UNPINNED
ssh root@<tv-host> 'printf %s "<movies-section-key>" > /tmp/plxnative-shared'
make run && tools/capture-screen.sh b-unpinned.png DISPLAY
```

`<movies-section-key>` is the `key` of a section that contributes shelves to Home —
`curl -H 'X-Plex-Token: …' http://<pms-host>:<pms-port>/library/sections`. On a stock layout the
movie library is `1` and TV is `2`; picking the TV one is the stronger capture, because Continue
Watching is usually drawn entirely from episodes and disappears whole (design §C's merge rule).

**A correct pair:** B differs from A by exactly that library's shelves — they are absent, with no
empty shelf, no spinner row and no placeholder (design §D). Everything else is byte-for-byte the
same screen: the tab strip still names the library, the profile chip, the clock and every other
shelf unmoved. **If anything else moves, that is the bug** — pinning governs Home only. Pressing
into that library's tab must still browse it normally, with its sort menu and A–Z rail intact.

**The log line that settles it either way**, in `/tmp/plxnative-events.log`:

```
pins: dev roster — 3 libraries, [1] treated as shared
pins: 2 of 3 libraries feed Home
```

Do **not** list every section: with nothing left pinned the rescue fires (Home is never empty by
design), so the screen is unchanged and correct, which reads exactly like a broken feature. The
second log line says `3 of 3` when that happens.

**Not reachable from a boot trigger, and host-tested instead:** the last-pinned refusal and the
persistence round trip both need the Sources list's OK, which lands in a sibling unit.


GLinnik21:

PR #34, opened 2026-08-13
strip: the tab projection needs a real type, or a friend's Music has no pill

GLinnik21:
Unit: **the top tab strip** (deliverable B).

## Which projection survived: #31's

I originally implemented the strip as a pure type roster — every pill a type name, so two of *your
own* movie libraries collapsed to one `Movies` pill. #31 landed first with the other reading of the
canvas: the strip names **your own libraries**, and a friend's library gets a pill only when it is a
type you do not own. **#31's projection survived**; I rebased onto it and deleted my duplicate
(`TypeRoster`, `pill_of_section`/`section_for_pill`, and the `/tmp/plxnative-sections` trigger, which
#31 makes redundant by registering `/tmp/plxnative-servers` into the registry for real).

Both readings are literally in the canvas — *"a pill is a type — not a library and not a person"*
versus *"It names the owner's own sections and nothing else — which is what it does today, so B is a
deliverable with no new drawing in it"* — and they coincide whenever a server has one library per
type, which is the case every frame in the design draws. #31's is the lower-risk one: no visible
change for existing single-server users, and no own library becoming unreachable when one server has
two of a type.

**This PR is therefore the delta on top of #31**, and it targets #31's branch so the diff is only
mine. Retarget to `shared-servers-base` once #31 merges.

## What this adds

**1. The projection needs a real type, or a friend's Music has no pill.** `is_show: bool` has two
values, so *"does an owned library have this KIND"* — the test that decides whether a borrowed
library gets its own pill — answers YES for a friend's **music** library on the strength of your
films. It folded onto your `Movies` pill and everything in it was unreachable from the strip. The
other half of the same gap was one layer up: `project_sections` dropped `artist`/`photo` at
discovery, so those types could never reach the projection to be *missing* from it. A type dropped
at the door cannot be the growth case #31 is written for.

`SecKind` (Movie/Show/Music/Photo) replaces the boolean, discovery keeps every type this product has
a level for, and the projection compares by type.

**2. The label + width cache moves onto the STRIP's generation** (`browse::tabs_gen`), bumped only
when the projected pill list actually changes. The section table's generation now moves once per
source as libraries land, and most of those fold onto pills already drawn — so the row re-measured
every pill, several times a boot, on Home's hot path, for a strip that had not moved.

**3. Two bugs of my own that review caught with probe tests** (fixed in `5531764c`), both newly
reachable *because* this branch admits `artist`/`photo`:

- **A profile switch kept the previous account's pills.** Keying the cache on the strip's generation
  made `reset()` stop invalidating it — `reset` *cleared* the projection's memo, so the next
  derivation compared `[]` against `[]`, saw no change, and never bumped. `draw_tab_row` iterates
  the cache, not the live table, so after a re-login the strip went on drawing *and hit-testing*
  libraries the new user cannot open. The memo is invalidated now, never emptied early.
- **The row grew by PEOPLE for an un-owned type.** `owned || !owned_kinds.contains` admits one pill
  per borrowed *library* of a missing type, so two friends who both share Music are two
  identically-titled `Music` pills. A missing type earns one pill now; the rest are reachable
  through the Source chip, exactly as a borrowed library of a type you *do* own already is. The
  invariance test could not see this because its fixtures were all types we own — it now grows that
  half too.

**4. A double free in `tabs()`, found by the new test.** The projection compared old against new with
`addr_of!(TABS).read()`, which copies the `Vec` **bitwise** and then drops the copy, freeing the
buffer the static still points at. It aborts the process (SIGTRAP, no panic message). Introduced by
me earlier in this change; it borrows now. Related and left alone: `tab_title`/`section_title` hand
back `&'static str` borrowed out of the table's own `String`s, which `append_sections` can
reallocate — sound for every caller in the app (they consume inside one frame with no append in
between) and unsound for a test that spans landings, which is why the new test owns its strings.

**5. Three comments said "the first section" where they now mean "the first tab"** (the navosc
scene, `Nav::select_pill`, the FPS manifest), and `nav.rs`'s `TAB` doc said the pill was a section
index plus one. Wording only — #31 wired the arithmetic correctly — but a doc that lies about an
index space is how the next edit gets it wrong.

## Tests

`make check` — **436 passed, three consecutive green runs**. `make` links; `make RELEASE=1` builds.

| test | pins |
|---|---|
| `the_strip_is_the_same_row_at_one_friend_and_at_three` | **the deliverable**: eleven libraries across four sources, and the pill list never moves. Width follows, since every pill's width is a pure function of its label |
| `a_friends_music_library_is_a_missing_type_and_not_a_second_films_pill` | the case the boolean could not express, plus `artist`/`photo` reaching the table at all |
| `only_a_changed_row_costs_the_tab_cache_a_re_measure` | the table's generation moves twice, the strip's does not; a gained pill does move it |
| `the_top_band_walks_the_projected_pills_not_the_section_table` | Home's focus packing clamps to the last PILL (5 libraries → 3 pills) |
| `a_profile_switch_re_measures_the_strip_instead_of_keeping_the_last_accounts_pills` | a reset must invalidate the row, not silently keep it |

Fixture names use the canonical stand-ins (`friend`, `nas-home`, `Film Club`) — character-length
preserving, for the width assertions in `home.rs`. `browse.rs` is fully scrubbed on this branch; the
other files still carrying the old names came from the base and are presumably covered by the
integration-branch scrub.

**Not verified on device.** I never touched the television: no deploy, no ssh, no capture.

The one behaviour a human should look at is **a library type that never had a tab before**. On an
account with a music or photo library, the strip gains exactly one pill for it:

```sh
tools/tv-session.sh up --screen home
tools/tv-session.sh shot /tmp/strip-home.png
tools/tv-session.sh key up right right right    # chip -> Home -> ... -> the new pill
tools/tv-session.sh key ok
tools/tv-session.sh shot /tmp/strip-music.png
```

CORRECT: one bare pill per own library plus any type only a friend has; the new pill opens that
library's grid, and the count read-out's noun matches the type (`films` / `shows` / `artists` /
`photos`). WRONG: a friend's music library sharing your `Movies` pill; a count saying `films` on a
music library; the strip re-measuring visibly (a pill flicker) as a second source's libraries land.

`./tests/run.py --fps` covers the two nav oscillators; `home-library-nav` is the one whose comment
changed here.

### Known gap, stated rather than hidden

A music library now has a tab and browses like any other section, but the level BELOW the grid does
not exist: OK on an artist opens the movie detail page, which has nothing to play. That belongs to
whoever builds the music level. The design's rejected case is a tab for a type with **no library**,
which is not this.

### DIFF
--- a/docs/pms-api.md
+++ b/docs/pms-api.md
@@ -32,8 +32,11 @@ Verified response (trimmed):
   {"key":"3","type":"artist","title":"Music",    "uuid":"332590ad-..."}]}}
 ```
 
-**Section keys on this server: Movies = `1`, TV Shows = `2`** (Music = `3`, ignore for now).
-Select sections by `Directory[].type == "movie" | "show"`, not by hard-coded key.
+**Section keys on this server: Movies = `1`, TV Shows = `2`, Music = `3`.** Select sections by
+`Directory[].type`, never by hard-coded key — and note the app no longer filters that type down to
+`movie`/`show`: `artist` and `photo` are kept too (`browse::SecKind`), because the top strip's
+projection can only give a friend's shared MUSIC library a pill if the type reaches the table at
+all. "Ignore music for now" was true while the strip named one server's movie and show sections.
 
 ---
 
--- a/rust-modules/src/app.rs
+++ b/rust-modules/src/app.rs
@@ -899,7 +899,9 @@ pub extern "C" fn plex_run(pms_host: *const c_char, pms_port: c_int) -> c_int {
             fn select_pill(&self) -> Option<usize> {
                 match self {
                     Nav::Home { .. } => Some(0),
-                    Nav::Library(sec) => Some(sec + 1),
+                    // a TAB index, not a section index: the strip is a projection of the table
+                    // (`browse::tabs`), so `+1` here is only the Home pill leading the row
+                    Nav::Library(tab) => Some(tab + 1),
                     Nav::Open { .. } | Nav::Back { .. } => None,
                 }
             }
@@ -3485,7 +3487,9 @@ pub extern "C" fn plex_run(pms_host: *const c_char, pms_port: c_int) -> c_int {
                     }
                     Route::Detail => nav_back(route, &trail, &mut nav_pending),
                     Route::Home => nav_to(route, Nav::Library(0), &mut nav_pending),
-                    // pill 1 is that same first section: Home comes back in the hero view with the
+                    // pill 1 is that same first TAB — not "the first section", which stopped being
+                    // the same thing when the strip became a projection of the table (`browse::tabs`):
+                    // several libraries can share one pill. Home comes back in the hero view with the
                     // top band on the pill the round trip started from, so the scene is a loop
                     Route::Library => nav_to(route, Nav::Home { focus_pill: Some(1) }, &mut nav_pending),
                     _ => {}
--- a/rust-modules/src/app.rs
+++ b/rust-modules/src/app.rs
@@ -872,7 +872,10 @@ pub extern "C" fn plex_run(pms_host: *const c_char, pms_port: c_int) -> c_int {
             /// always the Home pill). One word for both is why they were named apart: on the way
             /// back from the Library the selection moves to Home while focus stays on `Movies`.
             Home { focus_pill: Option<usize> },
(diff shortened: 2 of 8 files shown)


GLinnik21:

PR #35, opened 2026-08-13
library: a source that did not answer says so, and says whose it is

GLinnik21:
Deliverable **D** of the Shared Sources design — the section read-out when a borrowed server is gone, and the failed-sections-list layer above it.

## What this does

**`StatusOverlay` grows a reason line and an action.** Three blocks, each answering one question: the **verdict** (what happened), the **reason** (why, and what is *not* broken), the **action** (the one thing to press). Both optional blocks are ABSENT when unset — no gap, no empty band, no draw call — and everything hangs off one anchor, the caption band, in exactly the place it already occupied. So the player's and the person page's read-outs are laid out identically to before.

The caption is now a borrowed `&CStr` instead of `&'static`, because a section read-out interpolates a machine name and no `c"…"` literal can carry one. Every existing caller passes a `&'static` and is unchanged.

**Home's hand-built Retry is folded into the component.** It was a `Button::pill_w` and a y measured off the panel centre, sitting under a caption that came from the shared read-out — one block laid out by two pieces of code, agreeing only by inspection. The Library's read-out would have been a third copy. `action_frame` is now the one geometry the draw and the screen's hit test both read. Home's pill moves **7px up** onto the shared `space::LG` rung; that is the only intended visual change outside the Library.

**The dead-source read-out** is anchored in the **content region**, not at the read-out's canonical full-screen position — the chrome above it is still live, and that is the difference between a section failing and the app failing. Focus starts on *Try again*. Beside it: nothing. No Sort, no Filter, no item count, no A–Z rail (all four act on a grid with no items), no alert glyph (that mark is the player's), no BACK hint (BACK is universal; the chip and tabs above are the visible way out). `toolbar_n()` and `rail_drawn()` are one predicate each behind a control's draw, its pointer hit test *and* the focus walk, so neither control can be invisible-and-clickable or drawn-and-unreachable. When unit 7's Source chip lands, UP from the read-out reaches it with no change here.

**`StatusKind::Empty` keeps its no-danger-tint rule** and, on this screen, carries no action — the server answered, and asking it again gets the same answer.

**And the layer above.** `ensure_sections` swallowed every failure into `unwrap_or_default`, so a server that could not be reached at all was indistinguishable from a table nobody had asked for yet: no section, no state, `fetch_state()` answering `Loading` out of its `unwrap_or`, and the grid spinning for the rest of the session — the same bug the base branch fixed one layer down. `SECTIONS_FETCH` is that state; `fetch_table` is the one place "the server answered" and "the request failed" are told apart.

`readout_of` is where the whole decision lives now — one pure function over both fetch states and the store, replacing four `else if`s in the middle of `draw()` that no host test could reach.

## Two things found while reviewing this

- **Catching only the FAILED table was not enough.** A server that answers with nothing we browse (a music-only account) is `Ready` with zero sections and falls into the same two `unwrap_or` defaults — the eternal spinner, one case over. That one is `Empty`, because it answered.
- **…and its read-out was then drawn at alpha 0.** The grid's `Xfade` asked "did a section answer" to decide it had content; with no section that is false forever, so the fader never left its hold. `ready` now asks `readout()`, which is the question with the right shape.

`browse::failed_initial` is deleted — it was exactly one arm of `readout_of`, and one rule in two places is what this file keeps getting wrong.

## Host gate

- `make check` — **430 tests, green on three consecutive runs** (418 on the base), plus the clippy gate.
- `make` and `make RELEASE=1` both link for ARM.
- `code-review` (high) run twice — once on the first draft, once on the finished diff. Every finding in scope fixed: the zero-alpha `Empty`, the `failtest` trigger-name collision with the player, the rail focusable while undrawn, and the `Empty`-carries-no-action doc contradicting Home's *Refresh*. Two findings not in scope are noted at the bottom.

New host tests (10) cover the projection: failed-with-nothing shows the failure; failed-with-items does **not** blank a working grid; empty-and-reachable is `Empty` and never `Failed`; a failed sections list is observable rather than an eternal spinner; a table that answered with nothing browsable is `Empty`, not a spinner; the read-out is anchored below the toolbar; and the failure is drawn beside no grid control. `browse` gets three more for the table's own state machine.

**Not verified on device.** The TV was held for the whole of this work and the fleet device pass moved to the coordinator. Nothing below was run; it is written to be executed without me.

### Cheapest path: no proxy at all

**`/tmp/plxnative-libfail` forces the read-out directly** — added in this PR precisely because this screen's failure cannot be reached on purpose. It forces the *projection* only: `browse` keeps whatever state it really has, *Try again* still kicks a real fetch, and it is compiled out of `RELEASE=1` (verified — the string is absent from the release binary). It is deliberately **not** the player's existing `/tmp/plxnative-failtest`, which owns that name with a different vocabulary; arming that one used to pin this screen into a fake failure too, which the review caught.

```sh
# the failed BORROWED source — the design's case (deliverable D, figure 2)
ssh root@<tv> 'rm -f /tmp/plxnative-*; printf %s "<token>" > /tmp/plxnative-token;
               : > /tmp/plxnative-library; printf dead > /tmp/plxnative-libfail'
make run RUN_SECS=20 &
sleep 8 && tools/capture-screen.sh d-dead.png DISPLAY
```

Four variants, one per line of the design:

| content | screen |
|---|---|
| `dead` (or empty file) | borrowed source failed — verdict + reason + *Try again* |
| `own` | the same failure on your own server: verdict + *Try again*, **no** reason line (what every install draws today) |
| `table` | the sections-list failure, one layer up |
| `empty` | reachable and empty — a statement, no Retry, no danger tint |

### What each capture must show

- **`dead`** — three blocks, centred in the **content region** (top edge at the grid line, y=214, so the block's centre sits *below* the panel centre), not on the panel. Verdict `Can't reach <machine>` in the danger tint; reason `Shared by <handle> · your own server is fine.` one rung down in tertiary ink; one `Try again` pill, focused (accent fill, dark label). Above it: the profile chip and the tab strip, **undimmed and unchanged**. Beside it: **nothing** — no Sort chip, no Filter chip, no item count at the right margin, no A–Z rail, no alert glyph. Any of those appearing is a bug in this PR.
  - The reason line's `<handle>` will read **`a-friend`** — a documentation stand-in baked into this variant. Nothing maps a section to a source yet, so the product cannot produce that line at all today; the variant exists so the three-block layout can be *seen*. `<machine>` is the real session server name.
- **`own`** — same, minus the reason line, and the verdict sits at the frame centre with the pill one `space::LG` below it. This is the variant that ships today.
- **`table`** — same read-out; the tab strip above shows **only Home** (no section pills), because there is no table.
- **`empty`** — one line, tertiary ink, **no pill and no danger tint**, and no item count beside it.
- **Keys** (via `/tmp/plxnative-remote`, which the app drains every frame): `up` from the read-out must land on the tab strip (the toolbar has nothing to hold); `down` must come back; `ok` must not crash. `back` at the read-out goes to the tab row, then to Home.

### The honest path, if you want the real failure

`tools/netcond.py` between the TV and the PMS. Point the app at it by editing `PMS_PORT` in `src/config.local.h` and `make deploy` (host/port are compiled into `main.c`).

```sh
tools/netcond.py --listen <proxy-port> --target 127.0.0.1:32400 --control /tmp/netcond-d.mode
echo 'reject@/all?' > /tmp/netcond-d.mode     # the failed SECTION (the design's case)
echo 'reject'       > /tmp/netcond-d.mode     # the failed sections LIST, one layer up
```

- **Scope matters.** `reject@/all?` refuses only `/library/sections/<n>/all?…`, the page listing — so `/library/sections` still answers and the tab pills are real, which is what makes it a *section* failure rather than a dead app. Unscoped `reject` takes the sections list with it and gives you the `table` variant. Use `reject`, not `blackhole`/`stall`: the blocking paths here (`ensure_sections` on route entry, and *Try again*'s table branch) wait out the 2s connect / 15s `SO_RCVTIMEO` on a silent server, so the UI freezes rather than draws.
- **Pick a port Plex is not on.** It binds `127.0.0.1:32401` itself and the more specific bind wins, so the proxy is silently bypassed.
- **The macOS application firewall silently drops the TV's connections to an ad-hoc python listener.** netcond then logs nothing, which reads exactly like a quiet TV. The allow prompt must be accepted once per python binary path; start the proxy and confirm it logs a connection *before* going headless. (On this Mac `socketfilterfw --getglobalstate` currently reports the firewall **disabled**, so this may not bite — worth re-checking, since it is the failure that wastes an hour.)
- **netcond cannot produce the `empty` state at all** — it has no body rewriting, and that state needs a synthesized 200. `libfail=empty` is the only route to it.

### Then the gate

`./tests/run.py --fps` once, at the end. The scene that matters to this PR is `fps:home-idle`'s `fps_ceiling`: the failure read-out is **static by design** — no spinner, because a spinner for a source that isn't coming holds the panel presenting and gives back the measured 39.6% → 1.67% idle saving. Nothing here calls `ui::idle::invalidate()` on a clock; the state changes ride `browse::pump`'s existing landing invalidate and SDL key events.

## Not fixed here (out of this unit's files)

- `plex/probe.rs` ranks a hostname candidate ahead of a dotted quad, though `stream.rs` has no DNS — the harness's twin ranker in `tests/run.py` carries that tiebreak and the app-side one does not. Pre-landed policy with no caller yet; flagging for whoever owns `probe.rs`.
- A query the *server* rejects (a sort key it 400s on) fails the same way a dead server does, and the chips that could change it are not drawn — so *Try again* re-runs the losing query and the way out is another section or Home. Drawing the chips over the failure is in the design's rejected column, and clearing the user's filter behind their back is worse than either, so this stands; it is recorded at `toolbar_n`.

## Rebase note

Rebased onto `shared-servers-base` @ `c5ee486d`. Unit 7 (#31) touches `ui/library.rs` heavily — typed `Chip` list, `Nav::Library(n)` naming a pill — and is not yet in this base. Expect to rebase this on top of it; the collision points are `draw()`'s toolbar block, `move_focus`'s `Area::Toolbar` arm, and `enter()`.

---

## Merged with `shared-servers-integration` (2026-08-14)

This branch now **contains** the integration branch — unit 7 rewrote the ground it stood on, so the merge is a reconciliation, not a re-implementation. The PR's own base is still `shared-servers-base` only because `shared-servers-integration` is a local branch with no remote to retarget at; merge it from the local checkout. Against integration the diff is 5 files, ~980 insertions.

**Host gate after the merge: 510 tests green on three consecutive `make check` runs; `make` and `make RELEASE=1` both link; `libfail` still absent from the release binary.**

### `browse.rs` — my global goes, the per-source flags stay

`fetch_table` and `publish_table` are deleted outright. The integration branch's `ensure_sections` already carries the whole contribution — `found: Option<_>`, `ok = found.is_some()` — and records it **per source** as `reachable`/`sections_done`, which is strictly more than one global could say and is the same pair the Sources list dims a group by. Keeping `SECTIONS_FETCH` beside it would have been one rule in two places, which is what my own commit message argued against. `cur_source_state()` is a *projection* of those flags, so there is no fourth field to keep in sync.

`retry_cur_source()` moves the which-layer-failed decision into the module that owns the flags: an undiscovered table is re-fetched (blocking, the call route entry already makes, and it reaches the right server because `set_cur` moves `plex::current` with the browsed section), a failed page only clears its back-off.

### The one genuine design decision

**A source that stops answering keeps the sections it had learned and can still hold a page store full of items.** On the branch this came from, a failed table meant no sections at all — so "this source is gone" and "there is nothing of it on screen" were the same fact. They are not any more.

The read-out now takes the content region only when **both** are true: `total < 0` gates both of the source's terminal answers.

- **What it costs:** a grid you can still scroll but cannot play from — the posters are cached, the server is not there.
- **What the alternative costs:** a library blanked to report a server the Sources list is already dimming, and two layers of the same screen disagreeing about a rule the page layer has kept since before this branch ("a failed page over a populated grid leaves the grid alone").

I went with consistency. `an_unreachable_source_with_items_still_resident_keeps_its_grid` pins both halves.

### `ui/library.rs` — my predicates compose with the typed `Chip` row

`toolbar_n` is now `chips().len()`, and the whole suppression moved **into** `chips()`, which gained two arms: a one-source failure draws no chips at all, a multi-source failure draws `[Source]` only. That is the design's rule finally expressible — Sort and Filter are grid operations, the Source chip is **navigation** and is how you leave a dead source, which is exactly why the read-out spends no line telling you the way out. It could not be said before, because there was no Source chip to keep.

`focused_chip` returns `Option` for the empty row. Their draw loop already zeroes undrawn slots, so parked rects come free and their `w > 0.5` pointer scan is kept over my slice-by-length. `own_pill` keeps my clamp and takes their `tab_of_section` projection — a pill is a *type* now, and a failed discovery leaves zero section pills, reachable only because `enter` no longer bails before setting screen state.

### The merge made the reason line real

`dead_strs` had a hard-wired `None` and a note that nothing yet mapped a section to its source. `BrowseSource` **is** that mapping, so `browse::cur_source_labels` answers both halves and the design's three-block read-out now draws in the product rather than only under a trigger. An empty handle still means your own server, and the line is then *absent*, not empty.

The cache is keyed on the **strings** rather than on a generation: a machine names itself through `friendlyName` on a worker long after the table generation last moved, and the old key would have held `Can't reach ` forever.

### `ui/CLAUDE.md`

The coordinator's resolution taken as the base — it restores the `(server, path, w, h, png)` poster key and the `_on(ServerId, …)` twin, which this branch predates and would have reverted. Both bodies of prose are folded in: neither side's row described the other's half of the same module.

### Scrub

Seven identifiers, to the canonical stand-ins at the top of `docs/shared-servers.md`. Six were the ones unit 7's merge carried back in (the coordinator's `8ba098b1` fixed the same six independently — same target values, so the merge was clean); the seventh was in **prose this branch wrote**, quoting the design canvas verbatim for the empty read-out's caption. Worth noting as a class: prose written *from* the canvas carries the real values in as readily as a fixture copied from a capture does.


GLinnik21:

PR #37, opened 2026-08-14
person: a filmography is a fact about a person, not about one server

GLinnik21:
> **Base note — read first.** This unit branches from **`1dafd764` on `shared-servers-integration`**,
> which does not exist on `origin` (pushing it was refused in this session). The PR is therefore
> opened against `shared-servers-base`, the nearest pushed ancestor, so the diff shown here also
> carries the 56 commits between the two. **Mine are the last two — `cbae86fc` and `7fb15985`.**
> Files touched: `rust-modules/src/person.rs`, `rust-modules/src/ui/person.rs`,
> `rust-modules/src/pms.rs`, `rust-modules/src/plex/mod.rs`, `rust-modules/src/plex/servers.rs`.

## The bug

`person.rs` fetched a filmography with `plex::client_for(sid)` — **one server** — and presented the
result as *the person's* filmography. With a share registered that is a lie in both directions:

- a person opened from your own library shows only your films, silently;
- everything of theirs on a friend's server is invisible;
- and once the search screen can surface the same `actor` hub from a borrowed source, **which server
  you happened to arrive through decides what their page shows**.

It is the same page either way now.

## The join

**`tagKey`, and only `tagKey`.** A `personId` is server-local exactly like a `ratingKey`
(`docs/shared-servers.md` §1), so it addresses a *different person* on the next machine. Every
server but the origin therefore costs one extra round trip first — `/hubs/search?query=<name>`,
whose `actor` (then `director`) hub is joined back on the guid through the **existing**
`Tag::is_person`, not a second comparison. The hubs are scanned in **our** order, not the server's:
an actor tag and a director tag for one person are different rows with the same `tagKey`, and
`/hubs/search` reorders its hubs per query, so taking the director row would answer with only what
they *directed*.

The **origin skips the resolve entirely**: its credit row handed us its own id.

An **exact name** is the fallback, and only where there is no guid to join on — the entry carries no
`tagKey`, or we were opened without one. Never against a guid we can see and disagree with: a search
for a shared name returns both people, and a page of the wrong person's films is worse than a page
with fewer.

A server that has never heard of them **contributes nothing, and that is an answer** — it fails
nobody, shows no error, arms no backoff, and is never asked for a filmography it has no id for. It
also must not *settle* the page: `landed` folds as "any source has **content**, **or** every source
has answered". Both halves stop the same flash from a different side — a share that says "never
heard of them" instantly, and a `/media` call that succeeds with nothing shelvable (an
episodes-only credit lands exactly like that) must neither of them draw *"Nothing from this person
is in your libraries"* over the fetch still out on the server that does have them.

## Structure

`NFETCH` / `IN_FLIGHT` / `SLOT` / `RETRY_CD` become per-source over one flat index keyed on the
**registry slot**, never on a position in the source list — a slot is stable for the life of the
process, so a landing can never be applied to a different server because the list moved under it.
Three kinds per source (`K_RESOLVE` / `K_MEDIA` / `K_ROLES`) plus `F_PROFILE`, which stays **single**
because plex.tv answers about the person rather than about anybody's library.

Every discipline the file already had is kept: the monotone `land()`, `supersede()` as the one place
the generation moves and the mailboxes drop (walking the whole index space, so a slot that leaves the
roster cannot latch), the in-flight flag released when `spawn_small` refuses — and the client is now
**captured at the spawn site** (`pms::kick`'s rule) instead of resolved inside the worker.

The roles batch is filtered on the **source's own** local id. The origin's matches nothing on another
machine while still paying for the request.

## Three things that fell out

- **The page had no `ui::idle::invalidate()` at all.** A landing under a settled screen waited for a
  keypress to become visible. Every landing now repaints, failures included — the omission
  `search.rs`'s module doc warns the next store about.
- **The merged shelf divides its budget through `pms::allot`** instead of first-come. A prolific
  actor's 24 films on the first source would otherwise fill the row and starve every share behind it
  — this unit's own bug, one level down.
- **Focus follows the CARD, not the index.** Because `allot` re-divides the row on every landing, a
  bare column index silently comes to mean a different film when a slow share answers. `ui/person.rs`
  now snapshots the focused `(ServerId, ratingKey)` **pair** before the pump and re-seats onto it
  after.

## Deliberate non-goal

**Nothing is deduplicated across sources.** The only identity that could join two copies is the
portable `Metadata::guid`, which `PmsMovie` does not carry; a title+year join would merge two
different films *and* miss the case `docs/shared-servers.md` measured on this very pair of servers,
where the share's copy is titled in another language. A visible duplicate is honest; a wrong merge
hides a film.

## Blast radius

- `plex::MAX_SERVERS` — `pub(super)` → `pub`, for the reason `serverinfo` already imports it rather
  than restating 16: a store that guessed its own would go out of bounds the day that number moves.
- `pms::allot` — → `pub(crate)`, so there is one water-filling rule and not two.
- **`ui/person.rs` — 40 lines, all of them the focus re-seat** (`focused_id` + `reseat`, called from
  `update`) and one test. The shelf model itself needed no widening: `PmsMovie` already carries its
  `ServerId`, so `Art::Poster` resolves the artwork on the row's own server and `open_person_card`
  opens the card on the machine it came from, unchanged. The empty read-out ("Nothing from this
  person is in your libraries") reads *more* truthfully than before.

## Verification

- **`make check` — 545 tests, green 3/3** (and green 3/3 again after the review fixes).
- **`make` — the ARM cross-build links.**
- **Simulator, not the television** (the TV is in use). Booted to Home against the live PMS, then
  through the `plxnative-detail`/`detailsec`/`detailcol`/`detailok` chain onto a real person page and
  down into the shelves. `route=person`, header + portrait + `Movies 2` with the real total, both
  films, and on focus the character caption ("Thorin Oakenshield") — so `K_MEDIA` and `K_ROLES` both
  land through the new per-source path. Log: `person: source 0 'Richard Armitage' movies=2/2
  shows=0/0`.

**The multi-source path is NOT device-verified, and the simulator could not exercise it either.**
Only one server is reachable from this Mac (`plex: server slot 0 registered`, and nothing else), so
the cross-server resolve never ran for real. That half rests entirely on the host tests, which cover:
the `tagKey` join including a namesake the guid tells apart; the actor-director scan order against a
server that sent its `director` hub first; the exact-name fallback and its refusal against a
disagreeing guid; the merge keeping each row's `ServerId` and its parallel captions; `allot` beating
starvation; a source that contributes nothing without failing the others; both directions of the
`landed` fold; the focus re-seat across a re-divided shelf; and the stale-generation discard.

Also unverifiable here and worth a human with two servers: whether a real share's `/hubs/search`
`actor` hub carries the `tagKey` at all on every PMS version (this one does).


GLinnik21:

PR #38, opened 2026-08-14
search: the async store — one query, every source, merged into typed shelves

GLinnik21:
The fetch machine behind the types the seam commit (`1dafd764`) landed. `rust-modules/src/search.rs` only — no `ui/`, no `app.rs`, no `plex/`.

> **RETARGET THIS PR.** It is opened against `main` only because `shared-servers-integration` **does not exist on origin** — `gh pr create --base shared-servers-integration` fails with "Base ref must be a branch", and pushing someone else's branch is outside what this worker may do. Once that branch is pushed, retarget to it: **my own change is exactly one commit and one file**, `90b09516` / `rust-modules/src/search.rs`. Everything else in this diff is the unpushed `shared-servers-integration` history and is not mine to review. The other search units are pushed with no PR (`search-field-capsule`, `unit-5-search-recents`), so they are almost certainly stuck on the same thing.

**Base:** branched from `1dafd764` on `shared-servers-integration`; `git merge-base --is-ancestor 1dafd764 HEAD` passes, and local `shared-servers-integration` resolves to `1dafd764` exactly. Worth flagging for the fleet: my worktree was created at `05f286b3` (main), where `search.rs` and `ui/search/` do not exist at all — I reset the branch to `1dafd764` before starting. Anyone whose unit "can't find its file" has the same problem.

## What it does

`person.rs`'s idiom, not `browse.rs`'s: generation + a **monotone** mailbox write + `supersede` + a per-source in-flight claim, retry backoff and mailbox, pumped once a frame.

- **The monotone write is the whole reason `person.rs` was the model.** As-you-type guarantees overlapping workers — `wal`, `wall`, `wallace`, all in flight at once. `browse.rs` has no such guard, so a slow answer for `wal` would repopulate the results for a word the user finished typing two seconds ago.
- **Debounce** (`ui/detail.rs`'s `season_settle`, at 0.25 s): a five-letter word costs ONE round trip per source instead of four. Below `MIN_QUERY` nothing is asked at all — a one-character query returns every hub empty, so the first keystroke is free.
- **A trailing space repaints but does not re-ask.** The field draws the query verbatim; the server is asked the trimmed one. Collapsing the two re-asks the whole roster for identical terms every time the space bar is pressed, and drops an answer that is still correct.
- **Fan-out** over `plex::server_ids()`, each `ServerId` captured at the SPAWN SITE and resolved through `client_for` on the worker (`plex/CLAUDE.md`). A worker that read `client()` would search whichever server the user had wandered off to and file the answers under this slot's id, in another machine's `ratingKey` space.
- **A dead source contributes nothing and fails nobody.** Each has its own claim, backoff and mailbox; `State::Failed` needs EVERY source to have failed. One answer is `Ready`. An empty roster is `Failed`, because nothing can ever answer it and a spinner there would never end.
- `spawn_small` returning `false` releases the claim; `ui::idle::invalidate()` fires on **every** landing, failure branch included.

## The one real design call: the merge is ROUND ROBIN

This deliberately parts from how Home merges. Home groups shelves source by source because a shelf there BELONGS to a source and adjacency is what says so. A search shelf is genuinely mixed — `ui/search/results.rs` says the heading "cannot claim an owner", the annotation follows focus — so concatenating in roster order would bury a friend's best match behind twenty-three worse ones of ours. There is no cross-server relevance score to sort on; taking one from each source in turn is the most each server's own ranking can be honoured at once, and it makes the `MAX_ROW_ITEMS` cap fall on a round boundary rather than on one source.

Shelf order is `KINDS`, fixed, never the server's — the wire reranks its hubs per query (`sta` puts people first, `star` puts films first) and a row that moved under the user's focus mid-word is the bug that constant exists to prevent.

## Verification

- `cargo +nightly test --lib` — **547/547**, six consecutive parallel runs plus one `--test-threads=1` run. 12 of those are new.
- The new tests cover the hub→shelf mapping (both person hubs are one shelf; the wire's per-query hub order is discarded), the two payload containers (`Metadata[]` → `Item::Media`, `Directory[]` → `Item::Tag`, with tag id 0 kept as `""` rather than a person numbered zero), the round-robin merge and its cap, the four `State` verdicts, `set_query`'s transitions, the trailing-space rule, the debounce accumulator, a stale-generation landing being discarded while still releasing its claim, and a failed source backing off alone while its neighbour still answers.
- **Not yet run at the time of opening: `make` (ARM cross-build), `make lint`, and the simulator.** The machine was at 100% disk for most of this unit — most Bash calls died with `ENOSPC` before running — and the coordinator's standing order was to push before starting anything that rebuilds bundled FFmpeg per worktree. Everything in this file is `#[cfg]`-free Rust with no new FFI, so the cross-build risk is a warning at worst, but it should be treated as genuinely unverified until a follow-up comment says otherwise.
- One note for whoever reads a red suite: there is a **pre-existing** ~1-in-5 flake that aborts the whole test binary with a UTF-8/`char` precondition panic. I reproduced it on the untouched base with this file stashed out (535 tests, run 5 of 6). `--test-threads=1` never fails.


GLinnik21:
### All gates now green, and the store is device-verified against a real PMS

The body above was written mid-`ENOSPC` and said `make`, `make lint` and the simulator were unrun. They have all been run since, on `919ff1b7`.

| gate | result |
|---|---|
| `make check` (lint + host suite) | **550/550**, three consecutive runs |
| `make` (ARM cross-build) | links clean — `pkg/plxnative` produced |
| `make sim-shot` against the live PMS | the store fetched, merged and landed |

**The simulator run is the real evidence** — the screen draws nothing yet (other units own that), so the event log is the artefact:

```
search: q='wallace' settled, asking 1 source(s)
search: q='wallace' sid=0 asking limit=12
search: q='wallace' sid=0 hubs Movies=6 TV Shows=1 Episodes=12 Cast & Crew=3 Collections=1
search: q='wallace' shelves=5 items=23
search: q='wallace' state=ready
```

All five shelves populated from one live server, `Cast & Crew=3` proving the `actor` + `director` hubs merge into one shelf, and `Episodes=12` sitting exactly on `LIMIT` — which is what a per-hub limit looks like when it bites.

The `MIN_QUERY` gate, from a second run seeded with `w`:

```
loop=61 route=search fps=62 sim=1
```

No `search:` line at all. A one-character query costs no round trip, as documented.

### Review pass

`/code-review high` found four real bugs in this file, all now fixed in `919ff1b7`, all of which would have shown a wrong sentence on screen:

1. **`Ready` on the first answer printed "No results" over a source still out.** Our own server answers empty in 20 ms, a friend's takes a second — the user read "No results for wallace" and then watched a shelf appear under it. A pending source now holds `Searching`, unless some source has already given us something to draw (a populated screen is only added to, never contradicted).
2. **A late failure could un-say an answer.** The duplicate-spawn race `IN_FLIGHT` already admits to leaves two workers out at one generation; the loser's `None` flipped `Answered` → `Failed`, dropping already-drawn results and backing off for 2 s over a request whose answer was on screen.
3. **An empty roster spun forever.** `State` was only recomputed inside a landing, so the one case `state_from`'s empty arm exists for could never reach it — and it is reachable, since `/tmp/plxnative-search=<q>` forces the route whether or not a server was installed.
4. **`limit` is per HUB.** A response carries ~17 hub types, so 24 paid for up to ~400 records per settled keystroke on the endpoint whose whole point is being fast enough to type against. Now 12, which still fills the merged cap exactly from two sources.

### Findings for OTHER units (not mine to fix)

The same review pass turned up eight issues outside `search.rs`. Recording them here so they are not lost — each belongs to whichever unit owns the file:

- **`ui/search/mod.rs:157`** — the profile chip is drawn at a raw `54.0` instead of `widgets::CHIP_D` (60.0). The top strip is *continuous* chrome across the Home↔Search flip, so navigating visibly shrinks the avatar 6px and lifts its centre off the pill row's centreline. Also breaks `ui/CLAUDE.md`'s "never raw sizes".
- **`ui/search/mod.rs:145`** — the draw dispatch keys on `!q.trim().is_empty()` rather than `search::state()`. One character hides the recents and draws the empty state for a query the store never asked (`MIN_QUERY` is 2); and a real query in flight draws "No results" before any server has answered. `View` carries no `state` field, so the regions cannot correct this themselves — worth adding one.
- **`ui/search/mod.rs:136`** — `update` discards `pump`'s return value, which is documented as "so the caller can re-clamp focus". Once `move_focus`/`results::draw` are filled in, the shelf rebuild on each keystroke will leave focus indexing past the end.
- **`app.rs:1171`** — `leave_of(Route::Search)` is unreachable: `leave_of` is only consulted by `nav_back`, and Search leaves via `nav_to`/`nav_open`, both passing `leave: None`. So "the keyboard comes down with the page, at the fade floor" never happens; once `textinput` is real, opening a result leaves the system panel raised over the next screen.
- **`app.rs:3466`** — no `Route::Search` arm in the long-press repeat `match`, though `held_sym` is armed for it. Holding a D-pad direction on Search moves focus once and never repeats, unlike every other screen.
- **`app.rs:2904`** — no `Route::Search` arm in the mouse-down chain, so `ui::search::click`/`wheel` are defined and never called. Pointer *motion* was wired, so the pointer moves focus but a click does nothing — including on the tab pills the screen paints.
- **`ui/library.rs:791`** — `own_pill()`'s clamp `p.min(widgets::tab_count() - 1)` now lands on the *Search* pill rather than Home when the section table is empty, because the pill row grew by one. The comment above it ("With no section table there is only Home") is now false.
- **`plex/models.rs:167`** — `Metadata`'s struct doc was deleted by the `Hub::directory` hunk with nothing replacing it; looks like collateral rather than an intended removal.

GLinnik21:

PR #39, opened 2026-08-14
search: the /hubs/search query and its DTOs, pinned to a live capture

GLinnik21:
> ## ⚠️ RETARGET THIS PR before reading the diff
>
> **The base shown is wrong.** This branch is based on **`1dafd764`** ("search: the
> route, the pill and the seams"), the shared Search seam — which exists only on the
> local `shared-servers-integration` branch and **has not been pushed to origin**, so
> GitHub cannot use it as a base. `shared-servers-base` is 56 commits behind it.
>
> **Only the last 2 commits are mine.** Everything else in this diff belongs to other
> units. Once `shared-servers-integration` is pushed, retarget this PR to it and the
> diff collapses to the 3 files below.
>
> My unit's actual diff: `git diff 1dafd764..worktree-agent-a62b3264b3c16443d`
> — 3 files, ~440 insertions.
>
> (I tried to publish the base branch myself; the push was denied by the permission
> classifier, so it needs someone with that permission. Every other Search unit is
> blocked on the same missing ref.)

---

The `/hubs/search` query and its DTOs — the data layer under the Search screen.
Touches `plex/` and `docs/` only; nothing in `ui/`, `search.rs` or `app.rs`.

## What changed

`Client::search` existed and was dead. It now takes a `section_id`, both its numbers
are optional, and the path it builds is host-testable.

`Hub`/`Tag` needed **no new fields** — the shared base commit already carried
`Hub.directory`, `Tag.key`, `Tag.library_section_id`, `Tag.reason`/`reason_title`,
each with a lenient adapter. The audit is therefore spent as **tests against a
captured live response** rather than as new struct fields.

## Two findings the parameter names actively hide

Both probed against this household's PMS 1.43.3 on 2026-08-14.

**1. `sectionId` ranks; it does not filter.** `sectionId=1` moves the `movie` hub
ahead of `show`, `sectionId=2` puts `show` and `episode` first — and in every case
every row from every *other* section still comes back, same items, same counts. It
is a scope *hint* and cannot scope a search to one library. `plex-openapi.json` says
so too, in a sentence that is easy to read past: *"can result in re-ordering of
search result hubs."*

**2. A literal `0` on either number is an error the app cannot see.** This one is a
bug fix, not just documentation:

| request | result |
|---|---|
| `limit` omitted | **200** (3 rows per hub) |
| `limit=0` | **500** |
| `sectionId` omitted | **200** (whole server) |
| `sectionId=0` | **400** |

Every PMS error body on this endpoint is `text/html`, so `get_json` parses nothing
and returns `None` — the same `None` a dead socket gives. Written the obvious way
with `.int`, a search would **never work and would blame the network for it**. So
both numbers go on via `opt_int`, with a test that says why, so neither goes back.

(The brief asked only for `sectionId` via `opt_int`; `limit` turned out to need the
same treatment, which is only visible if you actually send `limit=0`.)

## Tests — 7 new, against a live capture

The substance of the unit. Five in `models.rs` pin what a future reader cannot
reconstruct:

- **The payload split** — `Metadata[]` for movie/show/episode, **`Directory[]` for
  actor/collection**. This is the failure that is silent and total: nothing errors,
  no field is missing, `Hub.size` still reports 3, and three of the five design
  shelves draw zero cards with a green parse.
- A person hit carries a portable `tagKey` and an absolute `metadata-static.plex.tv`
  thumb; a collection carries **neither, nor a `ratingKey`** — `key` is its only
  handle, so a screen keyed on `tagKey` silently drops every collection.
- **The same person arrives once per library section**, same `id` and `tagKey`, with
  a *per-section* count. Drawn raw he appears twice; keep the first row and you
  report 5 credits for someone who has 8. (Dedupe on `tagKey`/`id`, sum the counts.)
- Every number string-encoded — `score` arrives as `"0.52000"`, which is the live
  proof this endpoint does it.
- A matched-nothing query is an **answer**, not a failure: 17 hubs, all empty.

Two in `hubs.rs` cover the built path: percent-encoding of a user-typed query, and
the zero rule above.

## Fixture decision: inlined trimmed literal, not a `.json` file

Checked for precedent first — **there is none**: `find rust-modules -name '*.json'`
returns nothing, and every DTO test in `models.rs` inlines its body. Two reasons to
match that:

1. No fixture-file convention exists to follow, and the assertions read better next
   to the body they assert on.
2. A whole 40 KB capture would put one household's library — titles, people, section
   names — into a **public** repo for no test value. Trimming is what makes it
   publishable.

It is verbatim otherwise: **all 17 hubs in the server's own order, 9 of 19 rows
kept, every kept field and value byte-identical to the capture**, checked
mechanically rather than by eye (dropped fields, dropped rows, and the `size`
correction that follows a dropped row were the only sanctioned edits).

## Token check

- Token appears in **0 tracked files** and **0 diff hunks** (`git grep -F`, `git diff | grep -F`).
- PMS host appears in **0 diff hunks**.
- `src/config.local.h` confirmed gitignored (`.gitignore:25`).
- The only added `X-Plex-Token` string is the literal placeholder `X-Plex-Token=...`
  in the docs, matching the rest of that file.
- The fixture carries no token, host or `machineIdentifier`.

## Docs

`docs/pms-api.md` gains **§3b**, which it never had. The file's own banner used to
send readers to the spec for this endpoint — which is exactly where the wrong
example is — so the banner now says which half of the spec to trust.

## Code review

`/code-review high` ran against this diff. **Everything it found here is fixed** in
the second commit — including one genuine factual error I had propagated to six
places: three hub *types* arrive as `Directory[]`, but `search::Kind::hubs` folds
`actor` + `director` into one Cast & Crew shelf, so it is **two** shelves that
silently vanish, not three. It also caught an assertion that could not fail
(`is_person("0", "")` was run against a row whose id is 6068, so it passed
regardless; it now runs against `Tag::default()`, the only row that exercises that
guard).

Two findings deliberately **not** acted on, both pre-existing and outside this
unit's files-of-record — flagging rather than silently fixing:

- `Metadata.rating_key` is a bare `String` with no `de_str`.
- `MediaContainer.hub`, `Hub.metadata`, `Hub.directory` are strict `Vec` with only
  `#[serde(default)]`; a present `null` (as opposed to an absent key) would fail
  the whole container. `de_vec` exists for exactly this.

Both touch DTO fields every other unit depends on, the reviewer rated them low and
did not verify PMS ever sends `null` there. Follow-up, not a last-minute edit.

## Gates

Run in this worktree on 2026-08-14, at this branch's final content:

- `make check` — **542 host tests pass, 0 failed**; the three named clippy lints
  clean. (One transient failure mid-session, two clean runs after — a concurrent
  harness run, per `CLAUDE.md`'s warning about two jobs at once.)
- `make` — the ARM cross-build **links**; `pkg/plxnative` produced (8.5 MB).

No device or simulator run — this unit draws nothing. `tools/fwcompat.py` not run:
nothing here touches linkage or FFI.


GLinnik21:

PR #40, opened 2026-08-14
search: absolute headshots already reach the photo transcoder; collections have no art at all

GLinnik21:
> **Base note:** branched from `1dafd764` (the search batch's shared base on
> `shared-servers-integration`, which is not on origin), so this PR is opened against
> `shared-servers-base` like its sibling units. Only the last commit is mine —
> `rust-modules/src/posters.rs` and `rust-modules/src/search.rs`.

Artwork for search's two `Directory[]` result kinds. The headline is that the first half
needed **no new code**, and proving that is most of the value here.

## An `actor` thumb is absolute, and the existing path already handles it

A `/hubs/search` `actor` row's `thumb` is an absolute
`https://metadata-static.plex.tv/…jpg` — a different host, which `stream.rs` can never
dial (no DNS, no TLS). It does not have to. `Client::image_transcode_path` percent-encodes
whatever it is given into `url=` (`enc` → `pms::urlenc_str` is RFC3986-unreserved
passthrough, so `:` and `/` both become `%XX`), the request still goes to our own PMS, and
**the server does the TLS**.

**Live evidence, PMS 1.43.3, 2026-08-14** — status codes actually observed:

| request | result |
|---|---|
| `/photo/:/transcode?width=300&height=300&minSize=1&url=<percent-encoded metadata-static.plex.tv URL>` | **`200`, `content-type: image/jpeg`**, 12,318 bytes, `file` reports **exactly 300x300** |
| the same with the URL left RAW (unencoded) | `200 image/jpeg`, identical 12,318 bytes — so encoding is our discipline, not the server's requirement |
| the same with `url=` **empty** | **`404`, `content-type: text/html`**, 85 bytes |

So the transcoder does proxy third-party URLs on this server. There is deliberately **no
second code path** — relative and absolute differ only in how long the encoded value is —
and `poster_key`'s doc now records the route plus the rule `docs/pms-api.md` already
carried: never invent another image route for these.

## What did need fixing: an empty source must not become a request

`poster_key` used to turn an empty source into a fetchable-*looking*
`…&url=&X-Plex-Token=…` — the 404 above. Each one burns a slot as `P_FAILED` until the LRU
walks back to it. That was a rarity worth ignoring while every caller was
`ui::widgets::resolve_tex_on` (which refuses an empty path at the call site), but a
`collection` row carries no `thumb` at all, so search makes it the normal case for a whole
shelf. It now writes the empty key the store already reads as "nothing to fetch", and the
tile draws the skeleton face the design specifies.

The same gate covers a cliff absolute URLs brought within sight: `set_key` truncates into a
256-byte slot array, and a truncated key never equals the probe that built it — so every
frame would miss, claim a fresh slot and evict a real poster, the **whole store thrashing
over one tile**. A built path longer than a slot can hold is refused rather than handed
out. Nothing today is close (a headshot key measures ~177 bytes of the 255, a relative
thumb ~140) and the test derives the boundary from a probe build, so it keeps grading the
real edge if the request shape moves.

## Collections: skeleton, and here is the recipe if that changes

A collection row carries no `thumb` and no `ratingKey` — only `key`, a tag `id` and a
`collection://` guid — so the tile gets the skeleton face. **Recommendation: leave it
there**, but the resolve was measured and is written on `TagHit::thumb`:

- `GET /library/sections/{librarySectionID}/collections` → `Metadata[]` with real `thumb`
  paths, joined on **`index` == the search row's `id`** (or on `guid`). The `id` is **not**
  the collection's `ratingKey` — verified live, the two are different numbers.
- The cost is better than the brief assumed: **one request per library SECTION**, not per
  collection, and cacheable for the session.
- Both obvious shortcuts are dead ends — following `key` returns the collection's MEMBERS
  under the section's own generic art (`title2: "All Movies"`), and the
  `/library/sections/{k}/collection` tag axis carries no thumb either.

Still a second store with its own fetch, cache and invalidation for a shelf that is usually
a row or two, which is why the skeleton stands.

## Verification

- `make check` — **539 passed, 0 failed**, lint clean, including the four new tests in
  `posters.rs`: absolute URL percent-encoded into `url=` (and nothing raw surviving),
  relative key unchanged (plus the `format=png` clearLogo flavour), empty source yielding
  no request, and over-long paths refused rather than truncated.
- `make` — the **ARM cross-build links** (`pkg/plxnative`, ELF 32-bit ARM EABI5). It reused
  `vendor/ffmpeg-prefix` copied from the main checkout at the same commit rather than
  rebuilding bundled FFmpeg in this worktree; the `ffabi-assert` gate still compiled against
  those headers.
- No device run — the television was in use for this batch.

## One process note

The machine ran out of disk mid-run and this unit was blocked on it repeatedly. The cause
is structural rather than bad luck: `make` rebuilds bundled FFmpeg into **each worktree's
own** `vendor/`, so a fleet of parallel workers each doing that plus a ~1.2 GB cargo target
is arithmetic. Copying `vendor/ffmpeg-prefix` from the main checkout turns a multi-GB
two-minute build into a 4 MB copy, and deleting `rust-modules/target` on finish took this
machine from 118 MiB free back to tens of GB.


GLinnik21:
## Code review applied (commit `1a1cb99d`)

A review pass on the first commit turned up one finding that was a genuine placement error, so this is worth calling out rather than folding in silently.

**The length gate was in the wrong place.** `poster_key` refusing to hand out an over-long path only covers keys `poster_key` built — but `lookup` owns the 256-byte array, and it takes a `&str` from three `pub(crate)` entry points that each accept a raw `*const c_char`. A key built by any other route reached `set_key` ungated, got truncated, and could then never match its own probe: the whole-store thrash the gate exists to prevent. The check now lives at the array as `is_fetchable`, which also folds in the empty-key contract that was duplicated across `poster_get_wh` and `poster_warm`. Splitting it out is what makes it **testable** — `lookup` itself can't be reached from a host test binary (it drags in `gfx::delete_tex`, and nothing here links GL), so the precondition now has direct assertions instead of being demonstrated at one remove.

Also fixed:

- **Two ceilings, not one.** `cbuf::set` truncates to the *caller's* `cap - 1`, a different number from `KEY_MAX` and only in scope at `poster_key`. Both callers pass 352 bytes today, so this was latent — but the gate claimed a guarantee it didn't make.
- **A refused key was silent** — a permanent skeleton tile with nothing in `plxnative-events.log`, the whole support channel. Now logs once per process with both ceilings and the length that missed them; the latch matters because the memo makes a refusal repeat every frame. Same lesson `paths.rs` was fixed for.
- **`PT_KEYLEN` didn't finish its job** — two literal `256`s survived in the same file's docs, and the module doc still claimed ~135-byte keys, i.e. 40 bytes of headroom that no longer exist now a headshot key runs ~177.
- **Test hygiene:** the key tests mutate a second crate global, `KEY_MEMO`, which outlives `reset_servers_for_test`. They passed only because `Client::new` draws a process-unique `token_gen` — incidental. The fixture resets it explicitly now.

**Recorded rather than fixed:** `ui::widgets::tex_key` truncates the *source* path into its own 256-byte buffer with no gate, so two absolute URLs sharing a 255-byte prefix would collapse to one key. Unreachable today by arithmetic (such a source always builds past `KEY_MAX`, so it's refused anyway) — but that's an accident of three independent numbers, and `poster_key`'s doc now names which ones must not move.

**Three findings declined.** Two are pre-existing and outside this unit: `idle_of` treating `P_UPLOADING` as settled (real, but unreachable while the pump and every `store_idle` caller are main-thread), and the registry test fixture being triplicated across `plex/servers.rs`, `route.rs` and here — a shared `testlock`-adjacent fixture is the right answer, but it touches two other units' files mid-batch. One was **wrong**: `poster_key`'s `dst.is_null() || cap == 0` exit was reported as leaving a stale buffer unterminated, but `cbuf::set` opens with the identical guard, so the exit fires only in the two cases where writing even a NUL is impossible.

Gates re-run: **540 passed, 0 failed**, lint clean, ARM cross-build links.

GLinnik21:

Issue #41, opened 2026-08-14
cargo test --lib aborts ~1 run in 5 with a UTF-8 / char unsafe-precondition violation


GLinnik21:
## What happens

`cargo +nightly test --lib` aborts, roughly **1 run in 5**, with one of:

```
panicked at core/src/char/methods.rs:243:18:
unsafe precondition(s) violated: invalid value for `char`

panicked at core/src/str/mod.rs:435:9:
assertion failed: self.as_bytes()[0].is_utf8_char_boundary()

panicked at core/src/str/validations.rs:48:36 (and :55:40)

panicked at core/src/fmt/mod.rs:2949:38
```

always followed by `thread panicked while processing panic. aborting.` and
`process didn't exit successfully: … (signal: 6, SIGABRT)`. No backtrace survives — the
abort happens while the panic is being formatted.

Every signature is the same underlying event: **a `&str` being read while its bytes change.**
That is a real data race or use-after-free in the suite's process-global state, not a flaky
assertion.

## It is not the test that exposes it

Established on the **pristine base** (`1dafd764`), with `ui/search/empty.rs` reverted to its
untouched stub, by varying only what four added test bodies do:

| what the four tests do | runs | aborts |
|---|---|---|
| nothing added (bare base) | 20 | 0 |
| `assert_eq!(1, 1)` | 15 | 0 |
| `format!` in a 200-iteration loop, calling no project code | 12 | **2** |
| the real unit's tests (which allocate via `format!`) | ~40 | ~4 |

So **allocation churn from any new test is the trigger, not the cause**. The reading that fits:
a latent use-after-free is normally benign because nothing reuses the freed block; another
thread's `String` allocations reuse it, and the corruption becomes visible.

This matters for attribution — **a red `make check` may not be the author's change.** Re-run
before believing it.

## Narrowing so far

It needs tests from **both** halves of the suite to fire. Each of these was clean in 10 runs:

- `ui::` alone (280 tests)
- `stream:: browse:: auth:: task:: posters:: person:: metadata:: plex::probe ff::` (119 tests)
- `route::` alone (23 tests), 15 runs
- `ui::search::empty plex::` (58 tests), 8 runs

So it is cross-module: a global written by one module's tests and read by another's.

`--test-threads=1` **never** fails (539+ tests, many runs). That is the workaround, and it is
also the proof that this is concurrency and not a bad assertion.

## Where to look

[...]

PR #42, opened 2026-08-14
search: the two states with no shelves, as statements rather than faults

GLinnik21:
Unit 7 of the parallel Search build. **My commit is the single last one — `ad32711e`, one file,
`rust-modules/src/ui/search/empty.rs`.** `mod.rs`, the other `ui/search/` submodules, `search.rs`
and `app.rs` are untouched.

> **On the base:** this branches from `1dafd764` (the shared search seam) on
> `shared-servers-integration`, which is not pushed. Targeting `shared-servers-base` to match the
> sibling Search PRs (#39, #40), which means the diff shown here also carries the 56 commits
> between the two. Review `ad32711e` alone.

## What it draws

Four readings, decided from `search::state()` and **never from emptiness alone** — a `Ready`
store with nothing in it and a `Failed` one with nothing in it are the same store and opposite
meanings, which is the distinction `browse.rs` had to learn the hard way.

| state | drawn |
|---|---|
| `Idle`, empty field | "You haven't searched yet" |
| `Idle`, field holding letters | "Keep typing — 2 letters or more" |
| `Ready`, no shelves | `No results for "wallace"` (typographic quotes) |
| `Searching` | **nothing** |
| `Failed` | the app's centred failure read-out — see below |

Both statements are a `HEADLINE`-bold line in `TEXT_HEADING` over a `CAPTION`/`TEXT_TERTIARY`
hint one `space::SM` down, cap-top anchored at `CONTENT_TOP` on `MARGIN_X`, so the copy sits
where a first shelf's heading would. No alert mark, no danger tint, no Retry: nothing found is an
answer the server gave, and dressing it as an error tells the user something untrue about their
library.

## The two decisions the brief asked me to make and justify

**Plain copy, not `widgets::StatusOverlay`.** `StatusKind::Empty` is the right *reading* of
"nothing found" and the wrong *anatomy* for it here. `StatusOverlay` is the app's CENTRED
read-out: it centres its block on the frame, aligns every run `HAlign::Center`, and fixes its own
rungs (`BODY` verdict over a `CAPTION` reason, neither bold). This screen's copy is left-aligned
at the side margin on the `HEADLINE`-bold rung. Bending the component to that would mean adding
an alignment axis, an anchor mode and two size overrides to the one thing whose whole job is
holding the centre, and every existing caller would then be one builder call away from drifting
off it. That is a fork of its anatomy dressed as a style variant, which rule 4 of `ui/CLAUDE.md`
is aimed at.

Keeping the two anatomies apart also earns its keep: **a statement is page copy where the content
would have been; a fault is the centred read-out.** The design's own point is then made in
geometry, not only in tint.

**`State::Searching` draws nothing**, per the brief. A spinner belongs to whoever owns the fetch,
and a second unowned clock on one screen is worse than a quiet beat. Flagging a gap this opens,
for whoever takes the store unit: `set_query` clears the shelves on every keystroke, so between
the debounce firing and the answer landing the whole region below the field is currently
unowned and blank. Either the store should hold the previous shelves until the next answer lands,
or `results.rs` should own an in-flight read-out. It should not be added here.

**`State::Failed` is drawn, and reads as a fault.** It is here only because the routing above
sends it here: `mod.rs` hands this file every query whose shelf list is empty, and a failed
request has an empty shelf list — so an early return would blank the panel under a field holding
a query, which is exactly the "a screen that failed to draw" reading the not-yet copy exists to
prevent. It gets `StatusOverlay` at `StatusKind::Failed` (danger tint, no spinner) with a reason
line saying what is still fine, and **no action pill** — a control has to be focusable to be a
control and no `Zone` can hold one yet, so a pill the remote cannot reach would be worse than
none. Frame is clamped above `KEYBOARD_H` while the keyboard is up, per this screen's standing
rule. That arm is the thing to delete when the fault gets a real owner.

## Two rungs the design did not specify

The brief gives "across both sources" and "in &lt;server&gt;". Added:

- **&gt;2 sources → "across all sources."** The registry holds up to `MAX_SERVERS` = 16 slots, and
  "both" is simply false at three.
- **A source with no name → the sentence with no scope clause.** A boot that never reached
  plex.tv can dial a server and know no name for it (visible in the screenshots below — the
  simulator boots token-only, so nothing ever described the server). Never a placeholder name:
  saying less is better than inventing.

## Also not in the brief, and easy to drop if unwanted

- **"Keep typing — 2 letters or more"** for a sub-`MIN_QUERY` query. `set_query` maps anything
  under `MIN_QUERY` to `Idle` and clears the shelves, so backspacing a finished search down to
  one letter lands in `Idle` with the letters still on screen — the one place this file could
  state something the screen itself disproves ("you haven't searched yet", under the user's own
  typing). Caught by code review, not by me.
- **Typographic apostrophe (U+2019)** in "You haven't" and "Search didn't", not ASCII `'`. The
  brief asks for typographic quotes around the query; a straight mark beside a curly pair in the
  same type role reads as an accident. Font coverage verified.

## Verification — **simulator, not the television**

The TV was in use, so everything visual is `make sim-shot` on macOS at 1920x1080. Frame rates,
text rasterization and the video plane are all outside what this can answer; nothing here needs
them.

- `make check` — **540 tests green** (+5 mine). See the flake caveat below.
- `make` — the ARM cross-build links.
- `make lint` — clean, no new warnings.
- All four states photographed and looked at: empty field, sub-minimum query, no results, and the
  fault read-out.

The store stub lands nothing, so a real query never leaves `Searching`. The no-results and fault
shots were taken with a **temporary one-line fixture** in `say()` mapping `Searching` to the state
being photographed. **It is removed** — `say()` on this branch maps `Searching` to `None`, and the
test suite asserts it.

## Host tests (5)

The state table (including that `Ready`-with-shelves answers `None`, so a stray call cannot paint
"No results" over a screen full of them), the sub-minimum-query split, the typographic quoting and
that the closing quote survives eliding, and the scope rungs including the nameless and &gt;2 cases.
Placeholder server names only (`nas-home`) — this repository is public.

## Unrelated: the test suite has a memory-unsafety flake — #41

While verifying, `cargo test --lib` aborted several times with `invalid value for char` /
`is_utf8_char_boundary` / a panic inside `core::fmt` — a `&str` read while its bytes change.

**It is not this change.** On the pristine base with `empty.rs` reverted to its stub, four tests
whose bodies only `format!` in a loop reproduced it at the same rate (2 aborts in 12 runs), while
four `assert_eq!(1,1)` tests were 15/15 clean and the bare base 20/20. Allocation churn from any
new test is the trigger; the bug is a latent race that only becomes visible when another thread
reuses the freed block.

Consequence for anyone reviewing parallel work right now: **a red `make check` may not be the
author's change.** Re-run it, and `--test-threads=1` never fails. Full evidence and narrowing in
#41.


GLinnik21:

PR #43, opened 2026-08-14
search: the query capsule, its solid caret, and the scope line

GLinnik21:
Fills in `ui/search/field.rs`, which landed as a stub with a final signature.

**Read the diff for `rust-modules/src/ui/search/field.rs` only.** This branch is stacked on the search-seams commit `1dafd764` (`shared-servers-integration`), which is not published, so GitHub shows that base's commits here too. `git diff 1dafd764 --stat` on this branch is one file. `mod.rs`, the sibling submodules, `search.rs` and `app.rs` are byte-identical to the base — I verified that explicitly before committing, because I used throwaway local patches to `mod.rs` to reach states the boot trigger cannot (see below).

## What it draws

- **The capsule.** Pill radius, the card system's resting shadow (the same one every `widgets::Button` carries — without it an ACCENT capsule barely measures against its surround and only the dark label survives, floating). Focused *or* editing → `ACCENT` + `ACCENT_INK`; otherwise `CONTROL_IDLE_FILL` + `CONTROL_IDLE_INK`, so a held query stays legible without claiming focus it does not have. Those two states share one fill deliberately: the keyboard being up is the strongest statement that this control has the remote, and a field that dimmed while you typed into it would say the opposite. **No magnifier** — the pill above it is the mark.
- **The placeholder**, drawn whenever there is nothing readable in the field. `ROW_VALUE_INK_ON_DIM` over the accent fill, `TEXT_TERTIARY` when not.
- **The caret**: 3 × 34, solid, only while editing, and it must stay solid. `ui::idle` gates the whole frame present on motion it can detect and cannot see a clock, so a blink would either freeze on screen or hold the GL loop awake forever.
- **The scope line** at x=950 on the field's row, `CAPTION`/`TEXT_TERTIARY`, elided to the app's right margin.

Tokens and rungs throughout — no colour literal, no raw font size, every run placed by `Label`'s cap band.

## The two pure functions

Extracted and host-tested because each is the kind of thing that is wrong by a few pixels or one word and invisible in a screenshot.

**`run_layout`** — the field shows the **tail**. While the run fits it starts at the padding and the caret trails it; once it outgrows the box the whole run slides LEFT by the overflow, parking the caret's right edge exactly on the box's. A `Painter::clip` scissor (set and cleared inside the one draw, since it is global GL state) cuts the head at the padding instead of letting it slide out over the capsule's arc. The caret sits at the run's end with **no lead**, because that is the insertion point — the first cut had a 4px gap, which is visible as the bar jumping backwards on the first keystroke of every search.

**`scope_text`** — stated only when the scope is smaller than the account. All sources live it names them and attributes the share; one that did not answer names the machine and what you are looking at instead. No warning tint and no Retry: re-kicking a source is a control and it lives on that section's Library screen.

## Three boot-reachable states worth reviewing

These are the judgement calls, and the first cut got all three wrong in ways no screenshot would have caught. A code review pass found them; all three are fixed and each has a test.

1. **Ownership before the roster lands.** `plex::describe_server` is fed by plex.tv and by a server naming itself, both of which land *after* this screen's first frame — so `server_facts` is legitimately `None` at boot. Defaulting `owned` to `true` there labels a friend's machine "your server" (and, with two undescribed sources, reads "Searching your server and your server"). It now takes ownership from **registration order**, which is a documented invariant rather than a guess: `app.rs` registers shares after `plex::install`, "so slot 0 is always the session's own server".
2. **Attribution with two shares.** The handle is attributed only when there are two sources and **exactly one** is a handled share. A first-match pick would attribute one owner and silently drop the other into the same unbindable `·` the three-source rule already refuses.
3. **A whitespace-only query.** The screen below gates on the *trimmed* string, so a lone typed space must not leave this capsule blank and hintless while the rest of the screen says nothing has been asked. The query is also truncated at a NUL — `enter` is fed by a boot trigger that reads a **file**, so unlike keyboard input it is not guaranteed NUL-free, and `CString::new` failing would blank a field that still believes it holds text.

Not fixed here, and flagged rather than smuggled in: **the scope line has no invalidation source of its own.** This is the first per-frame draw to read `plex::server_facts`, and neither `plex::describe_server` nor the auth worker calls `ui::idle::invalidate`, so on a settled screen the roster landing is not presented until the 2 s keepalive. The right fix is one line at `describe`, which is a shared-module edit I did not want to make unilaterally with the rest of the shared-servers fleet in flight. It is written down at the call site. The fallback wording is what keeps the intervening frames a sentence rather than a hole.

Also left alone: two sources with the **same** machine name produce "nas-home unreachable · results from nas-home only". Real but narrow, and disambiguating it honestly needs more than this function has.

## Verification — simulator only, NOT device-verified

The television was in use, so everything below is `make sim-shot` on the macOS host. **Nothing here has been seen on a TV panel**, which means the rasterization and cap-band judgements this screen depends on are unverified where they matter most.

`make check` — **551 pass**, 16 of them in `ui::search::field` (6 on `run_layout`, 10 on `scope_text`). `make` cross-links clean for ARM.

Six states photographed and looked at:

1. **Empty, focused** — near-white accent capsule at 90,148,820×60 on a full pill radius; `Search` in the dim on-accent ink at the 32px padding. No magnifier in the field.
2. **`wallace`, focused** — the run in near-black on the accent fill, left-aligned at the padding, cap-band centred in the 60px row.
3. **`wallace`, editing, two sources** — the solid bar immediately after the final `e`, and the scope line reading *"Searching your server and Film Club · friend"*. That first phrase is the unnamed-source fallback doing its job: the simulator's own server has no roster name at boot, which is exactly the state finding (1) is about, and it produced a sentence rather than a hole.
4. **Empty, editing** — `| Search`: the caret at the insertion point with the placeholder clearing it, not sitting under it.
5. **`wallace`, focus elsewhere** — the dark idle capsule, query in white, no caret. Clearly unfocused, still legible.
6. **An overflowing run, editing** — the run has slid left, the head is cut mid-glyph on a clean vertical scissor edge at the padding, the tail is on screen, and the caret's right edge is parked exactly on the box's. This is the state the whole `run_layout`/scissor design exists for and it behaves as specified.

States 3–6 needed `mod.rs` patched locally (the boot trigger enters with `editing: false` and `zone: Field`, so the caret, the idle fill and the scope line are otherwise unreachable headlessly). Those patches were reverted before the commit and `git diff 1dafd764 -- mod.rs` is empty.

Server names, the library name and the handle are **placeholders** in the code, the tests and this description — `nas-home`, `Film Club`, `friend`. No real third-party data appears anywhere on this branch.

## One finding worth propagating (it cost several workers a hang today)

`make sim-shot` waits for `PLXNATIVE_SHOT_FRAME` **presented** frames, and `ui::idle` gates presents. A settled screen sits at `fps=0`–`1`, so the default `SIM_FRAME=200` needs ~200 *seconds* and reads as a hang with no output at all. Arm the DIAG trigger `plxnative-noidle` in `SIM_DIR` before shooting any screen that comes to rest — it is DIAG-exempt, so it does not also change which screen you boot to. This belongs in the `ui-sim` skill's recipe.


GLinnik21:

PR #44, opened 2026-08-14
search: the VKB seam — the TV's own keyboard, read at the offset this SDL uses

GLinnik21:
> **Reading this diff:** the branch is cut from `shared-servers-integration` @ `1dafd764` (the shared Search seams commit), which is not on the remote, so GitHub shows this against `shared-servers-base` and the diff carries that commit and the simulator work along with it. **This unit authored exactly two files and two commits** — `rust-modules/src/textinput.rs` and `rust-modules/src/app.rs`, `1dafd764..HEAD` — same shape as the sibling search PRs.

Fills in `rust-modules/src/textinput.rs`, which landed as a stub with a very good module doc, plus the two things in `app.rs` that make the events arrive at all. The app draws **no** keyboard of its own — the television draws it, over our surface, and hands the text back as ordinary SDL events. Nothing in `ui/search/` is touched.

## The offset is the whole job

LG's fork inserts a `Uint32 inputSource` at +12, so a committed string starts at **+16** on the television and at **+12** under `hostsim`, where desktop SDL2 is stock. Neither header in the repo says so: the vendored `include/SDL2/SDL_events.h` is stock 2.0.4 and states +12 as fact. A single hard-coded number does not fail loudly on the wrong platform — read a fork event at +12 and you land in `inputSource`, which is zero, so the field silently never fills; read a stock event at +16 and you commit the tail of the string.

So `TEXT_OFF` is a `cfg`, `decode_text_at` takes the offset as an argument, and the test asserts both layouts **and** asserts what reading each one with the other's offset produces.

Evidence, gathered before the code was written rather than after:

| Claim | Checked against |
|---|---|
| text at +16 on webOS | the NDK sysroot's fork `SDL_events.h` — `Uint32 inputSource;` then `char text[32]` |
| the mechanism exists on 4.x | `strings` over the real television `libSDL2-2.0.so.0.4.1` (SDL `2.0.5-83.gld4tv.15`) in the gitignored `sysroot/usr/lib/`: `text_model_interface`, `text_model_factory_interface`, `WebOSShowScreenKeyboard`, `WebOSHideScreenKeyboard` |
| plain link, not `dynlib!` | `tools/fwcompat.py --lib libSDL2-2.0.so.0 --grep TextInput` — the family on **all 14** images; same for `SDL_HasScreenKeyboardSupport` and `SDL_GetWindowFlags` |

`dynlib!` is for libraries whose SONAME moves. This is stock public SDL2 API, so there is nothing for a runtime bind to tolerate. The built ELF still grades **OK on 4.4.2 → 11.2.0**, unchanged; the pre-4.4.2 FAILs are the pre-existing `SDL_webOSCursorVisibility` and a Starfish symbol.

## `SDL_WINDOW_INPUT_FOCUS`, read for the first time in this tree

Trap 2 of the module doc: `SDL_StartTextInput` shows the panel only `if (SDL_GetKeyboardFocus())`, returns void either way, and **no window flag can ask for the bit** — the compositor grants it. Boot logs it beside `GL:`/`surface:`, and `start()` logs it again at the one moment it decides anything, because a boot reading is a false negative by construction (the flag arrives with the wayland keyboard `enter`, which needs the event loop that has not started yet).

Simulator boot line, verbatim:

```
keyboard: support=0 active=1 focus=0 winflags=0x26
```

`active=1` is not noise. SDL only auto-starts text input on platforms with **no** screen keyboard, so a desktop has had text events on since `SDL_VideoInit` while the television has them off until we ask. That is why the module tracks its own `STARTED` flag instead of `SDL_IsTextInputActive` — the latter would make `start()` a no-op on one of the two platforms — and why `start()` clears the pending buffer: without it, every keystroke used to navigate Home on the simulator would arrive as the first characters of the query.

## Do not fake the event with `SDL_PushEvent` — it SIGSEGVs

The obvious way to exercise this without a keyboard is to push a synthetic `SDL_TEXTINPUT`, exactly as `remote_synth_key`/`remote_synth_ptr` push synthetic keys and clicks. It killed the app twice before the backtrace named it, with **no Rust panic and no log line**:

```
EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x8
  libSDL3.0.dylib      SDL_PushEvent_REAL
  libSDL2-2.0.0.dylib  SDL_PushEvent_REAL
  plxnative-sim        app::plex_run::{{closure}}
  plxnative-sim        remote::Remote::drain
```

The Mac's `libSDL2` is **sdl2-compat forwarding into libSDL3**, and SDL3's text event carries a `char *text` **pointer** where SDL2's carries an inline `char text[32]`; the compat layer dereferences it while converting the pushed event. Keys and `ck:` clicks survive the same trip only because their fields are all scalars.

This is recorded at the call site and in the module doc, not only here — the next person to reach for synthetic events will hit it there.

## What is in the diff

- `textinput.rs` — `available`/`start`/`stop`/`drain` on the real SDL calls, the `cfg`-gated decoder, a bounded commit queue, `bind` for the focus probe, and `encode_event` (the decoder's inverse, coupled to it only by a round-trip test).
- `app.rs` — `SDL_TEXTINPUT` (0x303) and `SDL_WINDOW_INPUT_FOCUS`; the five SDL declarations in the existing `extern "C"` block; the ladder arm; the boot probe; the raw key dump widened 32 → 48 bytes (a TEXTINPUT payload ends at exactly +48, so the one event whose offsets are most easily wrong was the one whose forensic trail stopped just before its payload); and a `txt:` remote token.
- Second commit, from the code review: **the panel now comes down with the screen**. `leave_of`'s `Route::Search => search::leave` is consulted only by `nav_back`, and Search is left by `nav_to`/`nav_open` (a pill press, opening a result), which pass `leave: None` — so `stop()` was reached by no route off the screen. Per trap 3 the panel cannot be reopened once dismissed, so a panel left up over Home is a search field nobody can type into again all session. Fixed as a route invariant beside the screen's own per-frame update rather than by rewiring nav, which several other units are editing.

## Verification

| Gate | Result |
|---|---|
| `make check` | **546 pass**, 0 fail (13 new) + the three named clippy lints |
| `make` (ARM cross-build) | links |
| `tools/fwcompat.py` | **OK 4.4.2 → 11.2.0**, unchanged |
| simulator boot probe | `keyboard: support=0 active=1 focus=0 winflags=0x26` |
| simulator end-to-end | `txt:` through the real FIFO — see below |
| **device** | **NOT RUN — the television was in use for this unit. No claim is made about it.** |

End-to-end on the simulator, final build, tokens written to the real remote FIFO:

```
txt: decoded "wal" pending=1
txt: decoded "lace and gromit" pending=2
txt: decoded "Amélie" pending=3
txt: decoded "千と千尋" pending=4
txt: decoded "0123456789012345678901234567890" pending=5
```

`+` maps to a space (the FIFO protocol is whitespace-delimited, so a token can never contain one); multi-byte UTF-8 survives the raw byte field; a 40-character token truncates to 31 on a **char boundary**, not mid-codepoint; the queue accumulates because nothing drains it yet, which is correct while `ui/search/` is a stub. The app survived all of it, and the Home screen is pixel-unchanged.

**What the simulator does not prove**, stated plainly: no panel rises on macOS, so `SDL_StartTextInput`'s webOS side is unexercised; and because `txt:` deliberately bypasses `SDL_PushEvent` (see above), `SDL_PollEvent`'s own delivery of a real `SDL_TEXTINPUT` is untested. Only a real keystroke in `make sim-run`, or the television's panel, closes those. **The `+16` offset itself rests on the fork header and the device library strings, not on an observed event.**

## Host tests (13)

Both platforms' offsets and the cross-read corruption; NUL termination inside the fixed array; a full 32-byte commit with no terminator; invalid UTF-8 replaced rather than panicking (this runs inside the SDL event loop, where a panic takes the app down — `remote.rs` lost the app exactly that way); a short buffer; multi-byte round trips; encode/decode round trip; char-boundary truncation; queue order and drain-once; empty commits dropped; and the queue bound.

## Notes

- Nothing in `ui/search/` is touched, and no keyboard of our own is implemented.
- The disk filled to 0 bytes twice during this work (11 parallel worktrees); every gate above was re-run to green on the final build afterwards.


GLinnik21:

PR #45, opened 2026-08-14
search: the typed result shelves, on the shared strip

GLinnik21:
> **Read the last two commits, not the file list.** This branch forks from `1dafd764` on the local
> `shared-servers-integration`, which is not a remote branch — so GitHub diffs against
> `shared-servers-base` and shows the whole integration stack beneath it. **This unit's own change is
> `159431dc` + `0cc3a8c3`, and they touch exactly one file:
> `rust-modules/src/ui/search/results.rs`.** Every other file in the diff belongs to the base or to a
> sibling unit.

## What this is

Unit: **the typed result shelves** — `rust-modules/src/ui/search/results.rs`, and nothing else. One file changed.

Five shelves (Movies, TV Shows, Episodes, Cast & Crew, Collections) in `search::KINDS` order, which is **fixed**: the server reorders its hubs per query, and a shelf order that followed ranking would move the row under the user's focus while they are still typing. An empty type draws nothing at all.

It is `card_row::strip` throughout, not a sixth per-screen tile loop, so every treatment a poster wears elsewhere in the app travels here for free — the focus pop and its lifted shadow, the amber resume bar, the watched tick, and the off-axis cull that keeps `resolve_tex` off tiles nobody can see. The vertical half of that cull is this module's: a shelf whose whole block is off the panel never reaches `strip`, which matters because one query answers with five shelves against a 64-slot poster LRU.

## Things worth a reviewer's attention

- **The flow is authored; the lift is paint.** `stack_top` is the one stacker and the draw, the reveal rule and the hit-test all call it. The focused shelf's heading rises by `CardRow::lift()` — the shared clearance rule, `tile height × (focus_scale − 1) ÷ 2`, i.e. the design's 17px on a poster row — and moves nothing below it.
- **`title_lines` stays at ONE**, and this is the one place the search shelves cannot copy the person page. `LABEL_BLOCK` is 114 and fixed by the keyboard clearance the whole layout is built around; a two-line title makes `TileLabel::height` 122, which would put the caption 8px inside the next shelf's heading. A test grades that in both directions.
- **The owner annotation follows FOCUS, not the heading**, because a search shelf merges across sources and a heading cannot claim an owner. Focus can step straight from one borrowed source to another, and a point of ink cannot cross-fade between two words, so the run fades out, adopts the new handle while it is invisible, and rises again. With no source there is no run, no dot, no pad and no draw call.
- **The springs are advanced from the DRAW**, because a region's whole contract is `draw(Painter, &View)` and the state machine hands its regions no dt. `update(dt, &View)` exists and takes precedence the moment `super::update` calls it — **wiring that call at integration is the one thing this unit wants from the state-machine unit.**

Also exported for that unit, since they need this module's geometry and its per-row scroll springs: `shelf_top` / `row_top` / `content_h` / `reveal_shelf` and the pointer `tile_at`.

## Gates

| Gate | Result |
|---|---|
| `make check` (clippy's three named lints + `cargo +nightly test --lib`) | **green — 543 passed, 0 failed**, including 8 new |
| ARM cross-compile of the crate (`cargo +nightly build --release --target arm-unknown-linux-gnueabi`, the Makefile's flags) | **green**, no warnings from this file |
| Full `make` (links the ARM binary) | **not run** — see below |
| Simulator capture, opened and looked at | **done**, 4 states |
| Device (`tests/run.py`, `--fps`) | **not run** — the television is in use; the screen has no fetch and no input yet, so no scene can reach it |

**Why the full `make` did not run, plainly:** the machine ran completely out of disk mid-session (`ENOSPC`, ~20 agent worktrees × 1–2 GB of cargo artifacts), and clearing my own worktree took `vendor/ffmpeg-src` / `ffmpeg-build` with it. `make` would rebuild the bundled FFmpeg — minutes and gigabytes on a shared disk that other workers were still blocked on. The Rust half of the link is what this change can affect and it is green; the C link is untouched by a UI-only Rust change with no new FFI.

**Owed at integration, not done here:** `ui/CLAUDE.md` asks anything that animates for a device FPS scene (an `fps_floor` proving it still moves, an `fps_ceiling` proving it stops). `tests/manifest.json` has no `fps:search-*` scene and there is no oscillator trigger for this screen, because the screen has no store and no input yet. The host half of that obligation is met for the annotation spring.

## Verification: what the simulator showed

Four states, each opened and looked at. The store lands nothing yet (the fetch is another unit, in parallel), so the shelves were fed by a **temporary fixture in the worktree — built from the live home-hub catalog for real titles, posters and stills, plus fabricated people and collections — which is NOT in this branch**; `git status` is clean at the tip and the only file this PR touches is `results.rs`.

1. **Focus in the field** (the real boot state): every tile at rest, no caption, no annotation, and the caption band still reserved — the TV Shows heading sits at the identical y as in state 2, so nothing reflows when focus arrives.
2. **Focus on the Movies shelf**: `Movies  12 results · friend`, the focused poster popped with `Stardust` / `2007 · friend` centred beneath it, and the second shelf's heading carrying its count and no annotation.
3. **Episodes and Cast & Crew**: 420×236 landscape stills with their resume bars; the people shelf reads `5 people · friend` (`count_word` gets the noun right) over 250px circles wearing the person glyph, since fabricated tags have no headshot.
4. **Collections**: three skeleton posters — `Directory` entries carry no thumb, so that is the resting state and not a loading one — under `Collections  3 results · friend`, focused caption `Wallace & Gromit` / `6 items · friend`.

The owner annotation was reached through the existing `/tmp/plxnative-shared` trigger, the same file the hero's "Shared by …" run reads; the handle in every capture is the placeholder `friend`.

## Review

`/code-review high` was run and returned 12 findings. **Eight are fixed in the second commit**, one of which would have shipped a visibly broken screen: the present gate clears its motion flag at the top of the frame *before* it reads it, so a spring stepped in a draw is never heard and the focus pop crept forward one step per 2 s keepalive. The rest: a read-out keyed on item count alone (so a shelf position whose kind changed kept the wrong plural, and the scroll re-seat never fired on the store's clear), a caption naming the wrong person's server for the length of a fade, a collection wearing a different blank than the posters above it, a hit-test rebuilding the tile pitch from bare consts instead of the `RowStyle` the draw uses, per-frame `CString` allocation in the heading that defeated the baking beside it, and a misdocumented constant.

Three findings are answered deliberately rather than fixed, each with the reasoning at the call site: the episode tile draws the show's fanart because `pms::parse_item` discards the episode's own still, and the fix belongs in the store that parses these hits; and the `dev_source` reader and the source-annotation flow are acknowledged near-duplicates of `home.rs`'s, left un-consolidated because merging them means editing shared screen files during an eleven-way parallel merge — the note says where the consolidation belongs.


GLinnik21:

PR #46, opened 2026-08-14
search: the state machine — zones, the ▼ handoff, scroll and the keyboard

GLinnik21:
The Search screen's **state machine** — the spine the four region files (`field` / `recents` /
`results` / `empty`) draw against. Everything here is behind the stubs `1dafd764` landed;
**`rust-modules/src/ui/search/mod.rs` is the only file touched.**

> **Base:** targets `shared-servers-base` for the same reason #43 and #44 do — the wave's real base
> `shared-servers-integration` (`1dafd764`) is a local branch that was never pushed. Everything
> below `1dafd764` in this diff belongs to those already-open PRs; **review the two `search:`
> commits only** (`e7c05ce1`, `e931a23d`), which are one file between them.

## The ▼ handoff, which is the point of the unit

▼ off the field leaves it **only for something that is DRAWN**:

| query | terms remembered | shelves | ▼ lands on |
|---|---|---|---|
| — | — | — | **nowhere** — focus stays visibly put |
| — | yes | — | Recents |
| yes | — | yes | Results, shelf 0 |
| yes | — | no | **nowhere** — the "no results" statement holds no focus |

One expression answers it (`below_of`) and `draw`'s region dispatch reads the **same call**, so the
zone focus moves to and the region on screen can never be two different answers. Dropping focus
into an invisible list is a screen the remote has silently stopped addressing, which is the failure
nobody sees until a user presses DOWN twice on a blank page.

## Scroll

One `Spring`, so `ui::idle` hears the motion for free and this screen owes the frame gate no clock
of its own. The focused shelf's block top comes to `CONTENT_TOP`, clamped to the content's own end,
and it is **zero whenever the shelves do not hold focus** — forced rather than chosen: nothing clips
the flow, so a non-zero shift with focus on the field would draw shelf 0 up through the field and
the tab strip. That is also why ▼ lands on shelf **0** and not on the shelf last left — the shift is
already back at 0 by then, so restoring a deep row would jump the page under the user's eyes at the
exact moment they asked for a small move. The COLUMN is remembered and clamped, which is the half of
"lands where you were" that costs nothing.

A poster shelf's block comes out at `HEAD_TO_ROW + 375 + LABEL_BLOCK` = 549, which is
`consts::ROW_PITCH` **exactly** — a search shelf and a home shelf stack at one rhythm, and a test
asserts it rather than leaving the number a coincidence.

## What OK does on a COLLECTION — nothing, deliberately

Asked for explicitly, so stated plainly. A collection hit is a `Directory` row: **no `ratingKey`**,
so it is not a detail page, and what it *does* carry (`key` = `/library/sections/1/all?collection=…`)
is a filtered listing that `browse.rs` has no query for. There is no `Node` this could build and no
screen to mount it on, so `on_ok` reports `Action::None` and the tile is inert — rather than routing
somewhere that would be a lie about what was pressed. The shelf still earns its place, because "this
collection exists" is the answer a lot of searches are actually after. The arm is named and carries
the reasoning, so when browse learns a collection listing the next person lands in the right place.

The two that DO open: a media item → `Node::Detail { sid, rk, spot: default }`; a person →
`Node::Person { sid, key, guid, name, thumb }`, where `key` follows `metadata::Credit::person_key`'s
rule (the local numeric id, falling back to the `tagKey` guid) because the two ids address different
services and passing one for the other 404s. Nothing here calls `detail::open_rk` or touches the
route — the screen reports, `app.rs` performs.

## Region signatures: UNCHANGED

Confirmed, since four workers are building against them right now: **no region file was touched**,
and every `draw(p: Painter, v: &View)` signature is byte-identical to `1dafd764`. `View` gained no
fields and lost none — `shift` is now actually driven, and `recent`'s doc gained a paragraph making
the Clear index precise (`clear_index()` = one past the last term SHOWN, which is `MAX_RECENTS` only
once the list is full; on a two-term list it is 2, or ▼ would walk onto rows that were never drawn).

Two **additive** seams the regions may adopt, both inert until they do:
`note_recent_rect(i, rect)` and `note_tile_rect(row, col, rect)`. Pointer rects are recorded at
draw (`library::TOOL_RECTS`' idiom) and parked before each frame's regions run. They cannot be
derived from this side even in principle: a shelf's horizontal offset is the scroll spring inside
its own `card_row::CardRow`, which lives in `results.rs`. Until the regions record, the field and
the tab pills still hit-test exactly.

## Typing

`textinput::drain()` is drained every frame and appended through `search::set_query`;
`key(sym) -> bool` is the seam for backspace, which `app.rs`'s key arm has no route for yet (the
TV's keyboard delivers printable characters as `SDL_TEXTINPUT` but its delete as an ordinary key).
`String::pop`, not a byte, so a multi-byte term cannot be left invalid UTF-8. OK toggles the panel;
leaving the field for a zone below drops it **and records the term**; BACK drops it and records
**nothing** — BACK means "not that", and a list of the searches a user backed out of is a list of
their mistakes.

## Three things are pure, so a host can grade them

`library::readout_of`'s precedent. `below_of` — neither store behind it can be seeded on a host (the
recents list is the session file's, the shelves a live server answer), so a test of the live
function could only ever reach one of its four cases. `seat`/`step_results` — the ragged-grid
arithmetic that actually goes wrong, a column carried onto a shorter shelf. `scroll_target_for`.

## The review pass (`e931a23d`)

`/code-review` found six real bugs in the first commit; all six are fixed there, and four tests
came with them. The two worth knowing about because they were user-visible:

- **A click on blank ground activated whatever held focus.** `click` fell through to `on_ok()`
  unconditionally, so a stray click on the page background with the cursor resting on Clear **wiped
  the user's recent searches**. One `hit()` now serves hover and click, and a miss does nothing.
- **The pointer left the field without going through `leave_field`**, so `EDITING` stayed true
  under another zone — and that freezes the shelf scroll unconditionally, so ▼ walked focus off the
  bottom of the panel with nothing moving. The same failure this screen exists to prevent, reached
  through the other input path.

Plus: `focus_is_card` now asks `open_focused` itself (a collection is an item and opens nothing, so
the old test armed a press that dipped and committed nothing); `clamp_focus` stopped throwing the
remembered column away on every keystroke; the results hit test gained the `w > 0.5` guard its
recents branch already had; and `note_recent_rect` drops an out-of-range index instead of aliasing
it onto Clear's slot.

## Verification

- **553/553 host tests**, 18 of them new in `ui::search::tests` — the ▼ handoff in all four cases,
  the shrinking-shelf re-seat, the scroll target rule (both clamps and both frozen cases), the strip
  cursor's ends, the click-miss, the zero-size rect, the sub-`MIN_QUERY` query, backspace, the
  OK/BACK edges and `enter`'s mount. Every test touching a global holds both the module's own
  `ZLOCK` and `testlock::serial()`, and the animation carries the **two-sided** idle test
  `ui/CLAUDE.md` requires (reports while running, quiet at rest).
- **`make lint` clean**; **the ARM cross-build links** (`make` → `pkg/plxnative`).
- **Simulator** — the television is in use, so these are simulator results and nothing here has been
  seen on a television. Boots to `route=search` with the field seeded; keys move focus through every
  zone with no panic; the idle gate settles the screen (`loop≈46 fps≈0`) and wakes on each press
  (`fps` 15 → 50 → 61). A capture confirms the strip zone end to end: ▲ seats the cursor on the
  Search pill, ◀◀ walks the bright focus capsule to *Movies* while Search keeps its selected plate —
  i.e. `focused_pill()` is honest as the zone changes. Below the strip the screen is empty, because
  all four region files are still the stubs their own workers are filling; **this is evidence the
  state machine runs, not evidence of how the screen looks.**

## Two things that need someone else's file

Neither is fixable from `mod.rs` and both are wiring another unit owns:

1. **`app.rs` has no arm for a strip pill press.** ▲/◀/▶ walk the pills and the capsule travels, but
   OK does nothing: `on_ok()` reports `Action::None` there by design ("`app.rs` reads
   `focused_pill()` and performs the press") and the `Route::Search` OK arm only matches
   `Action::Open`. Compare `Route::Library`'s `Action::GoHome`/`GoSearch` arms.
2. **`key()`, `click()` and `wheel()` are not called from `app.rs` yet** — no `Route::Search` arm
   under the key fall-through, `SDL_MOUSEBUTTONDOWN` or `SDL_MOUSEWHEEL`. So backspace does not yet
   delete on the device. The functions are written and tested against; they need three call sites.

**Unrelated, but it will bite this whole wave:** the host suite has a **pre-existing parallel race**
— verified present at `1dafd764` with this file reverted to its stub, so it is not from this PR.
Roughly 1 run in 16 aborts with `SIGABRT` inside `str::validations` or `fmt`, i.e. a `&str` read
from a global while another thread writes it. `cargo test --lib -- --test-threads=1` is 553/553
every time. Worth someone owning, because it reads exactly like a new test being flaky.


GLinnik21:

PR #62, opened 2026-08-23
packaging: the launcher tile learns twelve languages, and the gate learns to see it (LG #41)

### LINKED ISSUE #41: cargo test --lib aborts ~1 run in 5 with a UTF-8 / char unsafe-precondition violation

GLinnik21:
Closes the metadata half of **LG App Self Checklist #41 (*Change Language*)**: the app now ships
`resources/<locale>/appinfo.json` for twelve languages, staged into the `.ipk`, gated on both
packaging paths, and documented with its sources and its unverified edges.

`docs/distribution.md` §12 is the full account. This body is the short version plus the three
things a reviewer should not have to dig for: what LG actually documents, what this change does
*not* close, and the two claims in the plan that turned out not to match the tree.

## What LG documents, and its tier

Two sources, neither of which is a television.

**LG, *App Localization* — `webostv.developer.lge.com/develop/guides/app-localization`.** This is
`develop/*`, so by this repo's own rule it is **web-runtime documentation: context, never proof for
a native app**. It states:

> To add app information in a specific locale, you need to first create a locale folder under the
> resources folder and then write an *appinfo.json* file under the correct location.

> In the *appinfo.json* file for localization, you should fill appDescription and title properties.

> the top-level *appinfo.json* file is loaded into the system memory of webOS TV first, and then
> other *appinfo.json* files are overridden depending on the locale setting.

> The app ID and version in each localized *appinfo.json* file are the same as those in the
> top-level *appinfo.json* file.

> If you use non-Latin letters (non-ASCII) in the *appinfo.json* file, you should save the file in
> UTF-8 without BOM format.

**webOS OSE, *appinfo.json* — `webosose.org/docs/guides/development/configuration-files/appinfo-json/`.**
The open-source platform's reference for the same descriptor read by the same SAM — not LG's web
runtime guide, and still not this firmware:

> To provide app information in a specific locale, you need to create locale folders under the
> `resources` folder, then `appinfo.json` files for each locale under correct locations

**The honest reading, stated as such in the doc.** The mechanism is documented; the *native*
applicability is **inferred**, not sourced. The strongest thing that can be said for it is that the
OSE page documenting the localization tree is the same page documenting `type: native` as a valid
application type, and that what reads the descriptor is SAM — which registers and displays an app
rather than running its code, and which neither page ties to the web runtime. That is a good
argument, not a measurement. **#41's metadata half is therefore implemented and gated, not
verified**, and `docs/lg-self-checklist.md` should say exactly that (see the note at the bottom).

## What ships

`pkg/resources/{de,es,fr,it,ja,ko,nl,pl,pt,ru,sv,tr}/appinfo.json`, each exactly two properties,
UTF-8 without BOM:

```json
{
  "title": "PlxNative",
  "appDescription": "비공식 네이티브 Plex 클라이언트 - Plex와 제휴 관계가 없습니다"
}
```

Three deliberate decisions, each with a gate behind it:

* **The title is not translated.** `PlxNative` is the brand `TRADEMARKS.md` reserves.
* **Exactly two properties.** A third would duplicate `pkg/appinfo.json` and put the version in a
  file `ci/bump-version.py` and the release tag guard have never heard of — `ci/flavor.py`'s
  "PATCH, DO NOT DUPLICATE" reached from the other side, and also what LG asks for.
* **Near-literal translations.** The English sentence is a **trademark disclaimer**; it is graded
  against the English rather than rewritten per market, and the mark stays Latin `Plex` in every
  locale (asserted — a transliteration would both weaken the disclaimer and misuse the mark).
  Chinese is deliberately absent: `zh` alone says neither Hans nor Hant, and guessing a script for
  a legal sentence is worse than shipping English.

## Packaging

**`ci/mkipk.py::stage_resources` stages the tree, not the Makefile** — `APP_FILES` is a flat
`cp … $(STAGE)/` that cannot express a two-level tree, and `make deploy` has no use for these
(a developer install's tile is not what QA looks at). This joins `packageinfo.json` and the real
`Installed-Size` as something the ipk carries and a scp dev loop never needed. **Consequence: the
localized descriptors are in the `.ipk` only**, so the device recipe below is an *install*.

**The flavour suffix is reapplied when staging** — the non-obvious half. A localized descriptor
carrying the bare tracked title would override `PlxNative debug` back to `PlxNative` the moment the
set was switched to Korean, restoring exactly the two-identical-tiles ambiguity `ci/flavor.py`
exists to prevent, in the locale nobody developing this app is looking at. The suffix is read off
`flavor.appinfo_for`'s **output** rather than re-derived, so one place still decides what a
flavoured title looks like; it raises rather than guesses if that rule stops being a suffix.

**Reproducibility re-measured, not assumed.** Two `mkipk.py` runs over one stage give one sha256,
checked for both flavours (`stage_resources` writes through `json.dumps`; `add_tree` normalises
identity, mode, mtime and order as before).

## Gates, proven to fail

Seven defects were planted and `ci/check-package.py` caught every one: a third property, a BOM, a
non-locale directory name, a translated title, a translation that dropped the mark, a tracked
locale that never reached the archive, and the tree deleted outright.

Three shapes worth calling out:

* the tracked half runs on **both** paths, including the no-package-staged branch `release.yml`'s
  `prepare` takes before a tag exists;
* the staged locale set is a **set equality** with the tracked one — a locale added and never
  shipped fails as loudly as a stale one left behind (the ipk-vs-deploy divergence that shipped a
  fontless package for months, caught on the way in);
* the payload is asserted **by path**. This is the load-bearing one — see the corrections below.

## Two claims in the plan that do not match the tree

Reported per the fleet rule rather than worked around silently.

1. **"`ci/check-package.py` pins the payload file set, so a new `resources/` tree fails the check
   until it is widened" — FALSE.** The assertion is `check(expected <= names, …)` where
   `names = {Path(m.name).name for m in members}` — a **subset** check over **basenames**. A
   `resources/ko/appinfo.json` contributes the basename `appinfo.json`, which the top-level
   descriptor already supplies, so the tree neither fails the check nor is seen by it. Nothing had
   to be widened; something had to be **added**, and it is a full-path assertion, because the
   existing one is structurally blind here — a resources tree that never reached the archive would
   have passed every other assertion in the file.
2. **"it asserts the staged file equals the tracked one byte for byte" — imprecise.** It is
   `appinfo == json.loads((ROOT / "pkg/appinfo.json").read_text())`, a parsed-dict equality, and
   only under `IS_STABLE`. Nothing about it constrains a `resources/` subtree, in either direction.

Neither changes the design; both changed which gate had to be written, which is why they are here.

## Two things the review pass caught, both fixed here

* **`mkipk.py`'s id guard was being defeated by the new staging.** The "staged `applications/<dir>`
  name is a fourth witness of the id" check ran at the *end* of `main()`, and `stage_resources`
  mkdirs `applications/<id>/resources/…` — so it created the very directory the guard reads.
  Reproduced against a wiped stage: the new code wrote a 1808-byte `.ipk` containing
  `packageinfo.json` and twelve locale files, **no binary, no top-level `appinfo.json`**, and
  exited 0, where the pre-change script exited 1. The guard now runs **before anything is
  written**, which is strictly better than where it was (a bad stage no longer produces an ipk at
  all) and is commented as load-bearing ordering rather than tidiness.
* **The build-machine-path scan labelled failures by basename.** There are now *thirteen* files
  called `appinfo.json` in the payload, so a real hit would have named a file nobody could
  identify. It labels by path within the payload now.

## Explicitly NOT closed

**Full UI i18n is out of scope.** ~203 display literals across 45 files, and `Label`/`Button` take
a non-owning `*const c_char` — an owned `String` needs a lifetime story a metadata change has no
business inventing. The CJK fallback face that landed separately is a *precondition* for a
translated UI, not the thing itself: glyph coverage means Korean renders, not that anything is
written in Korean.

**The app is locale-blind, and that is checkable rather than asserted:**
`grep -rn "setlocale\|LC_ALL\|localeInfo\|X-Plex-Language" rust-modules/src src` finds nothing. It
never reads the television's language and never sends `X-Plex-Language`, so PMS returns metadata in
the server's default language whatever the set is set to. That header is the cheapest remaining
i18n win by a wide margin and is a `plex/` change, not a packaging one — out of this unit's files.

## Device recipe that would settle #41

No television was touched by this lane, so none of this is verified. §12.4 has the full list; the
first item is the one that decides whether anything was closed:

1. `make FLAVOR=debug install` — an **install**, not a deploy: the descriptors are packaged, not
   scp'd. Then switch the set to 한국어 in *Settings → General → Language → Menu Language* and look
   at the launcher tile and the app-list entry. The tile should still read `PlxNative debug` (brand
   untranslated, flavour suffix survived); the description, wherever this firmware surfaces one,
   should be Korean.
2. If nothing changes, **reboot before concluding the file is unread** — webOS OSE's own docs warn
   SAM may cache `appinfo.json` at boot.
3. If `resources/ko/` is still not picked up, try `resources/ko/KR/` and `resources/ko-KR/` before
   concluding the mechanism does not apply to native apps. The two sources disagree about the
   folder shape and LG's example tree is an image.
4. Establish whether a webOS TV displays `appDescription` **at all** for a sideloaded app, or
   whether it is Content-Store-listing metadata only. If the latter, this closes #41's listing half
   and nothing visible on the set — still the right answer for a submission, but a different claim.

## One file this unit does not own

`docs/lg-self-checklist.md` §5 still reads *"That is the cheap half and it is not done."* — which
this change makes stale. It is not in this unit's file set (several lanes are closing checklist
rows into it), so it is left alone. Suggested replacement for whoever owns it:

> **#41 — the app is English-only, and the metadata half is now done.** N/A on the item's own
> precondition (no in-app language setting). The launcher tile and store listing carry a translated
> `appDescription` in twelve languages via `resources/<locale>/appinfo.json`, staged into the .ipk
> and gated in `ci/check-package.py` — `docs/distribution.md` §12, including the citation tier and
> the device recipe that would let this row be marked at all. **Unverified on a television**: no
> source in this repo proves webOS reads that path for a *native* app, so it is not markable yet.

## Verification

* `python3 ci/check-package.py` — green on the tracked path and, against a synthetic stage, on the
  packaged path (both flavours). *(No `make ipk`: it cross-compiles FFmpeg, and this lane's rules
  forbid it. The staged path was exercised by hand-staging a payload and running `ci/mkipk.py`
  directly — which is also what made the id-guard defect above reproducible.)*
* Seven planted defects, all caught (above).
* Reproducibility: two `mkipk.py` runs over one stage → identical sha256, per flavour.
* `make check` — 1062 host tests, lint, `ci/flavor.py --selftest`, `tests/test_harness.py`: all
  green.
* **No television.** No ssh, no deploy, no `tests/run.py`.

### DIFF
--- a/ci/check-package.py
+++ b/ci/check-package.py
@@ -13,7 +13,7 @@ import sys
 from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
-import flavor  # noqa: E402  — ci/flavor.py, the one descriptor transform
+import flavor  # noqa: E402  — ci/flavor.py, which DECIDES a flavour's id and title
 
 ROOT = Path(__file__).resolve().parent.parent
 FAILURES: list[str] = []
--- a/ci/check-package.py
+++ b/ci/check-package.py
@@ -31,6 +31,70 @@ def png_size(p: Path) -> tuple[int, int]:
     return struct.unpack(">II", p.read_bytes()[16:24])
 
 
+# ---- the localized descriptors ----------------------------------------------------------------
+#
+# `resources/<locale>/appinfo.json` gives the launcher tile and the store listing a title and
+# description per television language. LG documents exactly two properties in one — "In the
+# appinfo.json file for localization, you should fill appDescription and title properties. All
+# other properties are kept the same as the top-level appinfo.json file" — and requires UTF-8
+# WITHOUT BOM for non-Latin text. Both are gated below, and neither is cosmetic:
+#
+#   * a THIRD property in one of these files is `pkg/appinfo.json` duplicated. The version would
+#     then live in a file `ci/bump-version.py`, `release.yml`'s tag guard and every version
+#     assertion in THIS file have never heard of — the failure `ci/flavor.py`'s "PATCH, DO NOT
+#     DUPLICATE" exists to prevent, arrived at from the other direction.
+#   * a BOM makes the file unparseable to a strict JSON reader while looking identical in every
+#     editor and in `git diff`. LG asks for its absence by name, so this is their rule, not ours.
+#
+# This runs against the TRACKED tree in every invocation, including the one that has no package
+# staged — the tracked-only branch below is what `release.yml`'s `prepare` runs before a tag
+# exists, and a translation is exactly the kind of thing that gets edited by hand.
+LOCALE_RE = re.compile(r"^[a-z]{2,3}(-[A-Z][a-z]{3})?(-([A-Z]{2}|[0-9]{3}))?$")
+LOCALIZED_KEYS = {"title", "appDescription"}
+
+
+def check_tracked_resources(expect_title: str) -> list:
+    """Grade `pkg/resources/`. `expect_title` is the title these must carry. Returns the locales."""
+    src = ROOT / "pkg" / "resources"
+    if not src.is_dir():
+        check(False, "pkg/resources/ exists — the localized appinfo tree (LG checklist #41)")
+        return []
+    locales = sorted(p.name for p in src.iterdir() if p.is_dir())
+    check(bool(locales), f"pkg/resources/ carries at least one locale (saw {len(locales)})")
+    english = json.loads((ROOT / "pkg/appinfo.json").read_text())["appDescription"]
+    for loc in locales:
+        f = src / loc / "appinfo.json"
+        check(LOCALE_RE.fullmatch(loc) is not None,
+              f"pkg/resources/{loc} is a locale tag (language[-Script][-REGION])")
+        if not f.is_file():
+            check(False, f"pkg/resources/{loc}/appinfo.json exists")
+            continue
+        raw = f.read_bytes()
+        check(not raw.startswith(b"\xef\xbb\xbf"),
+              f"{loc}/appinfo.json is UTF-8 without BOM (LG's stated requirement)")
+        try:
+            d = json.loads(raw.decode("utf-8"))
+        except (UnicodeDecodeError, json.JSONDecodeError) as e:
+            check(False, f"{loc}/appinfo.json is valid UTF-8 JSON ({e})")
+            continue
+        check(set(d) == LOCALIZED_KEYS,
+              f"{loc}/appinfo.json carries exactly {sorted(LOCALIZED_KEYS)} (saw {sorted(d)})")
+        # The title is the BRAND, not a string to translate: TRADEMARKS.md reserves it, and LG's
+        # own listing shows it verbatim. A translated one here would also silently un-badge a
+        # flavoured install's tile, which is why `mkipk.stage_resources` reapplies the suffix
+        # rather than trusting whatever is written in the tree.
+        check(d.get("title") == expect_title,
+              f"{loc}/appinfo.json title is the untranslated brand ({expect_title!r})")
+        desc = d.get("appDescription", "")
+        check(bool(desc.strip()) and desc != english,
+              f"{loc}/appinfo.json appDescription is present and actually translated")
+        # The English sentence is a trademark disclaimer. A translation that transliterated the
+        # mark ("플렉스") would both lose the disclaimer's force and misuse it, and no reader of
+        # this repository is placed to catch that by eye in twelve languages.
+        check("Plex" in desc, f"{loc}/appinfo.json names Plex verbatim (the disclaimer's subject)")
+    return locales
+
+
 def font_family(p: Path) -> str:
     """nameID 1 out of a TrueType `name` table, without fontTools."""
     b = p.read_bytes()
@@ -106,6 +170,10 @@ if len(staged) != 1 and not REQUIRE_PACKAGE:
           f'pkg/appinfo.json id is the stable id ({flavor.STABLE_ID})')
     check(tracked_control.get("Package") == flavor.STABLE_ID,
           f'control Package is the stable id ({flavor.STABLE_ID})')
+    # Tracked too, and editable by hand in twelve languages — so graded on the cheap path as well,
+    # not only after a cross-build has produced a package.
+    print("== localized appinfo (tracked) ==")
+    check_tracked_resources(tracked_appinfo["title"])
     for f in FAILURES:
         print(f"::error::{f}")
     print(f"\n{'all tracked-file assertions passed' if not FAILURES else 'FAILURES above'}")
@@ -198,8 +266,12 @@ for member in sorted(PAYLOAD.rglob("*")) if PAYLOAD.is_dir() else []:
     if not member.is_file():
         continue
     dirty = sorted({m for m in HOSTPATH.findall(member.read_bytes()) if not ALLOWED_PATH.search(m)})
+    # Labelled by PATH inside the payload, not basename: since the localized descriptors landed
+    # there are THIRTEEN files called `appinfo.json` in here (the top level plus one per locale),
+    # and a basename cannot say which one is dirty — nor which twelve of the thirteen identical
+    # `ok` lines to stop reading.
     check(not dirty,
-          f"{member.name} carries no build-machine path"
+          f"{member.relative_to(PAYLOAD)} carries no build-machine path"
           + (f" (saw {dirty[0].decode(errors='replace')})" if dirty else ""))
 
 # ---- the RELEASE is coherent, not just the package -------------------------------------------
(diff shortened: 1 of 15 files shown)


GLinnik21:
## `doc-claim-auditor` pass — one finding lands outside this unit's files

Ran the auditor over the diff (73 tracked markdown files, the module docs of `ci/*.py`, the
Makefile's packaging comments, the workflow comments around `check-package.py`, and every Rust
module doc naming `appinfo.json`, `pkg/` or the app directory). **Two findings; one is fixed here
in 573faf8b, one is in a file this unit does not own.**

**Fixed (mine):** `ci/mkipk.py`'s module doc opened *"This module also SYNTHESISES **the two**
descriptors an ipk must carry … **both** derived from `pkg/appinfo.json` so the version stays
single-sourced"* — and I added a third bullet under it without touching the sentence that counts
them. Wrong three ways: it is three; the third is transformed from `pkg/resources/`, not derived
from `pkg/appinfo.json`; and "must carry" is false of it (no `packageinfo.json` registers nothing,
no localized descriptors registers an app that has lost its language). Rewritten.

**Not mine — `ci/flavor.py:2`.** Its headline still reads:

> Which install a package is for — the ONE transform from the tracked descriptors to a flavour's.

`stage_resources` is now a second application of that rule, to twelve more tracked descriptors. The
single point of *decision* survives — it string-diffs `flavor.appinfo_for`'s **output** rather than
re-deriving the suffix, and raises `SystemExit` naming `ci/flavor.py` if the rule ever stops being a
suffix — but the uniqueness claim as written does not. The two echoes of that line inside this PR's
own files (`ci/mkipk.py`, `ci/check-package.py`) are retired in 573faf8b; the headline itself is
left for whoever owns that file. Suggested replacement:

> Which install a package is for — the one place that DECIDES what a flavour's id and title are.
> `ci/mkipk.py::stage_resources` applies the title rule a second time, to the localized
> descriptors, reading the suffix off this module's output rather than re-deriving it — so the
> decision is here and the application is in two places.

### Two pre-existing staleness items the pass surfaced (not caused by this change, not touched)

* `README.md:229` — *"Notably the app links the TV's own FFmpeg and GLib **dynamically** rather
  than bundling them."* Inverted for FFmpeg since bundling landed: the ipk ships `lib*-plx.so.*`,
  and the Makefile's own `APP_FILES` comment says shipping it "makes that our obligation rather
  than the television's". This is the public readme, in the licence section.
* `ci/flavor.py:11` and `docs/two-installs.md:148` both say `pkg/appinfo.json` has **fourteen**
  fields. It has fifteen, and their enumeration of "the other twelve" omits `appDescription` — the
  field this PR makes load-bearing.

### Verified still true (i.e. nothing else to chase)

Root `CLAUDE.md`'s `make ipk` and `make deploy` bullets and its `pkg/` key-files list;
`docs/two-installs.md` §2's three id gates and §3.1's *"The launcher tile — its own `title`"*
(which the flavour-suffix reapplication exists to keep true); `docs/distribution.md` §9 and §11;
the Makefile's `APP_FILES` basename comment; `.claude/skills/cut-release/SKILL.md`'s four-file
version claim (the localized files carry no version, so it survives); `which-tier`'s packaging
paragraph; `tests/README.md`; and the five Rust module docs that mention `pkg/appinfo.json` or the
app dir — none describes the app directory as flat or enumerates its contents.

PR #69, opened 2026-09-06
fix(plex): dial a custom access URL's advertised port, not the PMS de…

frank-m:
…fault

A shared server can advertise a custom server-access URL — `https://host` with no port — whose real port lives only in the connection's `port` field (443, a reverse proxy). `candidates()` pushed the `uri` verbatim, so `Origin::parse` applied its 32400 PMS default and the probe dialled a closed port: the whole server then read as unreachable, and a server with no relay connection had no fallback to mask it. `plex.direct` URIs are unaffected — they always spell their port.

Add `Origin::parse_connection(uri, advertised)`, which falls back to the connection's advertised port only when the URI omits one (a port the URI wrote still wins, and a written-but-garbage port is still refused), and use it when building the URI candidate. Regression-tested from the real `/api/v2/resources` shape, watched red on the verbatim-uri behaviour first.

---------------------

This PR was made with claude.

It fixes a small issue that excludes some remote servers from being loaded.

Awesome work btw, really like the client :)

### DIFF
--- a/rust-modules/src/plex/origin.rs
+++ b/rust-modules/src/plex/origin.rs
@@ -161,6 +161,30 @@ impl Origin {
             .then(|| Origin::new(p.scheme, p.host, p.port.unwrap_or(DEFAULT_PORT)))
     }
 
+    /// Parse an advertised `Connection.uri`, but fall back to the connection's own advertised
+    /// `port` — not [`DEFAULT_PORT`] — when the URI wrote none.
+    ///
+    /// plex.tv's `plex.direct` URIs always spell their port (`…plex.direct:32400`). A **custom
+    /// server-access URL** does not: `https://plex.example.com` carries no port, and its owner is
+    /// serving it on 443 (a reverse proxy), which is exactly what the connection's `port` field
+    /// reports. Parsing such a URI through [`parse`] would apply the 32400 PMS default and dial a
+    /// closed port — the whole server then reads as unreachable, with no relay to fall back on
+    /// (observed live: a share whose only remote was `https://<custom>` at `port:443`, timing out
+    /// at every probe deadline while the official client reached it fine). A URI that DID write a
+    /// port keeps it; `advertised` is the fallback, and only a `port_bad` URI (a written-but-
+    /// garbage port) is still refused.
+    pub fn parse_connection(uri: &str, advertised: i64) -> Option<Origin> {
+        let p = Parts::of(uri);
+        if !p.scheme_known || p.host.is_empty() || p.port_bad {
+            return None;
+        }
+        let port = p
+            .port
+            .or_else(|| dial_port(advertised))
+            .unwrap_or(DEFAULT_PORT);
+        Some(Origin::new(p.scheme, p.host, port))
+    }
+
     pub fn scheme(&self) -> Scheme {
         self.scheme
     }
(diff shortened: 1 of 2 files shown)


GLinnik21:
> Awesome work btw, really like the client :)

🥰

rust-modules/src/plex/probe.rs
```diff
@@ -731,6 +737,53 @@ mod tests {
...
 
+    /// A **custom server-access URL** advertises its `uri` WITHOUT a port (`https://<host>`) and its
+    /// real port only in the connection's `port` field (443, a reverse proxy). The candidate must
+    /// dial that 443, not the 32400 PMS default `Origin::parse` applies to a portless URL — the bug
+    /// that made a real share (only remote a custom `https://` at `port:443`, no relay) unreachable
+    /// while the official client reached it. A `plex.direct` URI that already spells its port is
+    /// unaffected.
+    #[test]
+    fn a_custom_access_urls_port_comes_from_the_connection_not_the_pms_default() {
+        // Peetoom's shape, anonymised: one unmatched-LAN plex.direct + one portless custom remote.
```
GLinnik21: nit: let's not name people here, public repo :) "a reporter's shape" is fine

rust-modules/src/plex/probe.rs
```diff
@@ -345,7 +345,13 @@ pub fn candidates(res: &Resource) -> Vec<Candidate> {
         if !c.uri.is_empty() {
             let scheme = scheme_of(&c.uri, &c.protocol);
             if !(unmatched_shared_lan && scheme == Scheme::Http) {
-                push(c.uri.trim_end_matches('/').to_string(), scheme);
+                // Not `c.uri` verbatim: a custom server-access URL (`https://host`, no port) must
```
GLinnik21, line 350: the doc comment above `candidates()` still says "verbatim" — mind updating it too?

rust-modules/src/plex/origin.rs
```diff
@@ -161,6 +161,30 @@ impl Origin {
...
+    /// plex.tv's `plex.direct` URIs always spell their port (`…plex.direct:32400`). A **custom
+    /// server-access URL** does not: `https://plex.example.com` carries no port, and its owner is
+    /// serving it on 443 (a reverse proxy), which is exactly what the connection's `port` field
+    /// reports. Parsing such a URI through [`parse`] would apply the 32400 PMS default and dial a
+    /// closed port — the whole server then reads as unreachable, with no relay to fall back on
+    /// (observed live: a share whose only remote was `https://<custom>` at `port:443`, timing out
+    /// at every probe deadline while the official client reached it fine). A URI that DID write a
+    /// port keeps it; `advertised` is the fallback, and only a `port_bad` URI (a written-but-
+    /// garbage port) is still refused.
+    pub fn parse_connection(uri: &str, advertised: i32) -> Option<Origin> {
```
GLinnik21: optional: `advertised: i64` would save the `as i32` round trip at the call site

GLinnik21 (approved):
Thanks for this, and for the kind words. Nice catch, and a clean fix.

GLinnik21 (changes requested):
Thanks again for this! Just requesting changes for the couple of small things I mentioned above. Otherwise looks good to me.

frank-m:
fixed the findings and used squash/force push to get rid of the name.

GLinnik21 (approved):

PR #78, opened 2026-09-10
fix: register Original MDE before the Part GET

MrcRjs:
PMS 1.43 503s an Original Part GET whose session never registered a Media Decision Engine handshake (`Denying access due to session lacking permission to direct play`). Smart direct-play used to skip `/decision` so a TrueHD/DTS default would not veto an AAC/AC3/EAC3 sibling, which is exactly the path that then 503'd.

## What this PR changes

Every Original plan now asks `GET /video/:/transcode/universal/decision?hasMDE=1&directPlay=1` first. The query names the track Original will actually play:

- `audioStreamID` is the chosen AAC/AC3/EAC3 sibling (smart-DP), so MDE does not evaluate TrueHD/DTS and force a video-downscaling transcode.
- `subtitleStreamID` is always sent. An advertised embedded codec Original client-renders (PGS, mov_text, dvd_subtitle, …) is named; a sidecar or unadvertised codec is `0` so MDE evaluates subs off instead of burning.

If `/decision` is unreachable or unusable, the plan fails closed: it does not return the Part URL. Remux or re-encode can still register via a separate `transcode_decision`.

`Part.decision=transcode` is a container change, not a video-copy veto. A remux that copies video is forbidden only when the **video stream's** own decision is `transcode` (bit depth, an undecodable codec). The measured TrueHD-only and Profile 5 shapes are `Part=transcode` with video `copy`; treating the part as a re-encode would transcode 4K for an audio problem. Remote Auto's Original probe then samples that remux `start.mkv` rather than the Part (a Part GET after a transcode decision is 503).

The direct-play profile's `subtitleCodec=` list is the same `DP_SUBTITLE_CODECS` constant the MDE stream-id gate reads, so a selected PGS cannot be advertised in one and omitted from the other.

Empty `session=` stop/ping/close calls no longer hit the wire (PMS logs those as "without a valid session GUID").

`tools/plxnative-lab` treats a missing multicast route as "UPnP not found" instead of crashing the selftest offline.

## Verification

Host tests in `route.rs` / `plex/{models,transcoder}.rs` cover:

- HEVC+EAC3 registers MDE before returning the Part
- smart-DP names the AC3 sibling on `/decision`
- `Part.decision=transcode` with no video stream remuxes (does not GET the Part)
- TrueHD-only remuxes; a video-stream `transcode` forbids remux; Profile 5 still refuses a remux copy
- Remote Auto probes `start.mkv` after a transcode MDE, never the Part
- unreachable/empty MDE does not return the Part URL
- selected PGS / mov_text / dvd_subtitle named; sidecar and unadvertised codecs send `0`

### DIFF
--- a/docs/plex-api-catalog.md
+++ b/docs/plex-api-catalog.md
@@ -302,7 +302,10 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
   id, or `0` so a selected sidecar or unadvertised codec does not force a burn). If
   `/decision` is unreachable the app does **not**
   return the Part URL (that would 503); it remuxes or re-encodes instead. An explicit MDE
-  `transcode` also refuses a local codec-copy remux. (Catalog **D-7** still describes
+  `transcode` still remuxes when the video stream's own decision is `copy` or unnamed
+  (`Part.decision` is a container change); only a video-stream `transcode` forbids the
+  codec-copy. Remote Auto's Original probe then samples that remux `start.mkv` rather than
+  the Part (a Part GET after a transcode decision is 503). (Catalog **D-7** still describes
   `audioStreamID`/`subtitleStreamID` on *transcode* start/decision GETs as undocumented relative
   to the PUT selection API; the MDE handshake above is a separate use of those query names.)
 - **Divergences:** none (the key comes straight from `part.key`). The middle path segment is a
--- a/docs/plex-api-catalog.md
+++ b/docs/plex-api-catalog.md
@@ -224,9 +224,9 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
 
 - **Response:** no body (200) / 400 if the stream doesn't belong to the part.
 - **App reads:** only the HTTP status. Confirmed.
-- **Divergences:** none. This is the **correct** mechanism for stream selection (the app also
-  redundantly appends `audioStreamID`/`subtitleStreamID` to the transcode GET — those are
-  undocumented no-ops there; see D-7). `partId` is an integer in the spec — keep parsing it as
+- **Divergences:** none. This is the **correct** mechanism for stream selection. Transcode
+  start/decision GETs still copy the same ids (they match this PUT); MDE GETs use them as
+  load-bearing query names — **D-7**. `partId` is an integer in the spec — keep parsing it as
   numeric.
 
 ### 9. `transcodeDecision` — register/handshake a transcode session
@@ -270,9 +270,13 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
     **query params**. The spec declares these as **HTTP headers** (`X-Plex-Client-Identifier`
     is a **required header**). PMS accepts X-Plex-* as either, so it works, but header form is
     spec-correct and avoids leaking identity into cached URLs.
-  - **D-7:** `audioStreamID` / `subtitleStreamID` on the transcode GET are **not documented**
-    params of `transcodeDecision`/`transcodeStart`; server-side selection is done via
-    `libraryPutPartsPart` (which the app already does). The GET copies are no-ops — safe to drop.
+  - **D-7:** `audioStreamID` / `subtitleStreamID` are **not documented** params of
+    `transcodeDecision`/`transcodeStart`. Their meaning splits on `directPlay`:
+    - **MDE** (`hasMDE=1`, `directPlay=1`): both are load-bearing (smart-DP sibling as
+      `audioStreamID`; `subtitleStreamID=0` so a sidecar or unadvertised codec does not burn).
+      Not safe to drop.
+    - **Transcode** start/decision (`directPlay=0`): PUT `/library/parts/{id}`
+      (`libraryPutPartsPart`) remains the selection API; GET copies match that PUT.
 
 ### 10. `transcodeStart` — the actual transcoded MKV stream
 - **Method / canonical path:** `GET /{transcodeType}/:/transcode/universal/start.*`
@@ -305,9 +309,8 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
   `transcode` still remuxes when the video stream's own decision is `copy` or unnamed
   (`Part.decision` is a container change); only a video-stream `transcode` forbids the
   codec-copy. Remote Auto's Original probe then samples that remux `start.mkv` rather than
-  the Part (a Part GET after a transcode decision is 503). (Catalog **D-7** still describes
-  `audioStreamID`/`subtitleStreamID` on *transcode* start/decision GETs as undocumented relative
-  to the PUT selection API; the MDE handshake above is a separate use of those query names.)
+  the Part (a Part GET after a transcode decision is 503). See **D-7** for the MDE vs transcode
+  split on those query names.
 - **Divergences:** none (the key comes straight from `part.key`). The middle path segment is a
   **changestamp** (part updatedAt), not a byte offset — the app treats it as opaque, which is
   correct.
@@ -386,7 +389,7 @@ These documented operations cover things the app currently improvises or doesn't
 | **D-6a** | transcode decision/start | `maxVideoBitrate` → spec name `videoBitrate` (or `peakBitrate`). | low |
 | **D-6b** | transcode decision/start | `session=` → spec name `transcodeSessionId` (+ `X-Plex-Session-Identifier` header). | low |
 | **D-6c** | transcode decision/start | `X-Plex-*` sent as query params; spec declares them as **headers** (`X-Plex-Client-Identifier` required). | low |
-| **D-7** | transcode decision/start | `audioStreamID`/`subtitleStreamID` on the GET are undocumented no-ops; correct mechanism is `libraryPutPartsPart` (already used). Drop the GET copies. | low |
+| **D-7** | transcode decision/start | Undocumented `audioStreamID`/`subtitleStreamID` on the GET. **MDE** (`hasMDE=1`, `directPlay=1`): load-bearing (smart-DP sibling; `subtitleStreamID=0` so a sidecar does not burn) — do not drop. **Transcode** (`directPlay=0`): PUT `libraryPutPartsPart` is the selection API; GET copies match that PUT. | low |
 | **D-8a** | `/:/timeline` | App uses **GET**; spec verb is **POST** (`timelinePostSlash`). | medium |
 | **D-8b** | `/:/timeline` | `X-Plex-Client-Identifier` sent as query; spec = required **header**. | low |
 | **D-9** | `/video/:/transcode/universal/stop` | Undocumented endpoint (works); no spec drop-in. Nearest: `statusPostTerminate`. | low |
--- a/docs/plex-api-catalog.md
+++ b/docs/plex-api-catalog.md
@@ -247,7 +247,7 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
   | `videoResolution` | query | no | string | `1920x1080` | ✅ |
   | `offset` | query | no | number | seconds, on seek/retranscode | ✅ |
   | `subtitleSize` | query | no | integer | `100` | ✅ |
-  | `subtitles` | query | no | enum(auto/burn/none/sidecar/embedded/segmented/unknown) | `burn` | ✅ |
+  | `subtitles` | query | no | enum(auto/burn/none/sidecar/embedded/segmented/unknown) | `burn` on transcode; `none` on MDE | ✅ |
   | `videoBitrate` / `peakBitrate` | query | no | integer | — | app instead sends `maxVideoBitrate` (D-6a) |
   | `transcodeSessionId` | query | no | string | — | app instead sends `session` (D-6b) |
   | `audioStreamID` / `subtitleStreamID` | — | — | — | app sends them | **not documented on this op** (D-7) |
@@ -274,7 +274,8 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
     `transcodeDecision`/`transcodeStart`. Their meaning splits on `directPlay`:
     - **MDE** (`hasMDE=1`, `directPlay=1`): both are load-bearing (smart-DP sibling as
       `audioStreamID`; `subtitleStreamID=0` so a sidecar or unadvertised codec does not burn).
-      Not safe to drop.
+      Always `subtitles=none` (client-rendered). Omitting it leaves PMS on `auto`, which 1.43.4
+      HTTP 400s when the part already has a selected subtitle. Not safe to drop.
     - **Transcode** start/decision (`directPlay=0`): PUT `/library/parts/{id}`
       (`libraryPutPartsPart`) remains the selection API; GET copies match that PUT.
 
@@ -303,7 +304,9 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
   `GET /video/:/transcode/universal/decision?hasMDE=1&directPlay=1` first so the session is
   admitted; smart-DP names the chosen AAC/AC3/EAC3 track as `audioStreamID` on that query, and
   the decision always carries `subtitleStreamID` (an advertised embedded client-rendered
-  id, or `0` so a selected sidecar or unadvertised codec does not force a burn). If
+  id, or `0` so a selected sidecar or unadvertised codec does not force a burn) and
+  `subtitles=none` (PMS 1.43.4 400s `hasMDE`+`directPlay` with a selected subtitle and
+  the default `auto`). If
   `/decision` is unreachable the app does **not**
   return the Part URL (that would 503); it remuxes or re-encodes instead. An explicit MDE
   `transcode` still remuxes when the video stream's own decision is `copy` or unnamed
@@ -389,7 +392,7 @@ These documented operations cover things the app currently improvises or doesn't
 | **D-6a** | transcode decision/start | `maxVideoBitrate` → spec name `videoBitrate` (or `peakBitrate`). | low |
 | **D-6b** | transcode decision/start | `session=` → spec name `transcodeSessionId` (+ `X-Plex-Session-Identifier` header). | low |
 | **D-6c** | transcode decision/start | `X-Plex-*` sent as query params; spec declares them as **headers** (`X-Plex-Client-Identifier` required). | low |
-| **D-7** | transcode decision/start | Undocumented `audioStreamID`/`subtitleStreamID` on the GET. **MDE** (`hasMDE=1`, `directPlay=1`): load-bearing (smart-DP sibling; `subtitleStreamID=0` so a sidecar does not burn) — do not drop. **Transcode** (`directPlay=0`): PUT `libraryPutPartsPart` is the selection API; GET copies match that PUT. | low |
+| **D-7** | transcode decision/start | Undocumented `audioStreamID`/`subtitleStreamID` on the GET. **MDE** (`hasMDE=1`, `directPlay=1`): load-bearing (smart-DP sibling; `subtitleStreamID=0` so a sidecar does not burn; `subtitles=none` so a selected subtitle does not 400) — do not drop. **Transcode** (`directPlay=0`): PUT `libraryPutPartsPart` is the selection API; GET copies match that PUT. | low |
 | **D-8a** | `/:/timeline` | App uses **GET**; spec verb is **POST** (`timelinePostSlash`). | medium |
 | **D-8b** | `/:/timeline` | `X-Plex-Client-Identifier` sent as query; spec = required **header**. | low |
 | **D-9** | `/video/:/transcode/universal/stop` | Undocumented endpoint (works); no spec drop-in. Nearest: `statusPostTerminate`. | low |
--- a/docs/plex-api-catalog.md
+++ b/docs/plex-api-catalog.md
@@ -295,9 +295,16 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
   a decision and no decision could be inferred" — i.e. some server configs require a
   `transcodeDecision` call even for direct-play. PMS 1.43 logs that 503 as "Denying access due
   to session lacking permission to direct play" (or "due to terminated session" after
-  `/stop?closeResourceSession=1`). Every Original Part GET now registers
+  `/stop?closeResourceSession=1`). Every Original Part GET registers
   `GET /video/:/transcode/universal/decision?hasMDE=1&directPlay=1` first so the session is
-  admitted; smart-DP names the chosen AAC/AC3/EAC3 track as `audioStreamID` on that query.
+  admitted; smart-DP names the chosen AAC/AC3/EAC3 track as `audioStreamID` on that query, and
+  the decision always carries `subtitleStreamID` (an advertised embedded client-rendered
+  id, or `0` so a selected sidecar or unadvertised codec does not force a burn). If
+  `/decision` is unreachable the app does **not**
+  return the Part URL (that would 503); it remuxes or re-encodes instead. An explicit MDE
+  `transcode` also refuses a local codec-copy remux. (Catalog **D-7** still describes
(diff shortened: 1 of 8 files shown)


rust-modules/src/route.rs
```diff
@@ -5894,7 +5956,24 @@ fn build_stream(rk: &str, part: &str, vcodec: &str, acodec: &str, env: &ResolveE
...
+                    } else {
+                        // Part GET 503s after a transcode MDE. Sample the remux we would actually play.
+                        measure_remote_remux(
+                            client,
+                            rk,
+                            &session,
+                            env.audio_sid,
+                            env.sub_sid,
+                            source_transport_kbps,
+                        )
```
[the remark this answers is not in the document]
MrcRjs: Agreed this runtime climb still Part GETs. That is a different identity problem than this cold-start probe: `probe_original_while_hls` has to reuse the live HLS encoder and must not mint a /decision on that id. 

Recovery apply already remuxes via `candidate.direct`. Follow-up is a child remux sample that leaves HLS selected. Out of scope here.

rust-modules/src/route.rs
```diff
@@ -5061,23 +5084,38 @@ fn flavors_allowed(
...
+    audio_stream_id: i64,
+    subtitle_stream_id: i64,
+) -> Option<MdeVerdict> {
+    let mc = match c.mde_decision(rk, session, audio_stream_id, subtitle_stream_id) {
         Some(mc) => mc,
         None => {
-            // failed fetch OR unparseable (XML/truncated) body — keep the fallback visible
-            // in the event log, like the old raw-body scan did
-            crate::player::log("decision: no/unparseable response -> local heuristic");
+            // failed fetch OR unparseable (XML/truncated) body — no Original Part
```
GLinnik21: this fails closed on a timeout or unparseable body too, no retry, so one flaky MDE call pushes the whole library to transcode; keep the old heuristic for transport-class failures and let "server said no" be the only hard refusal?

rust-modules/src/plex/transcoder.rs
```diff
@@ -194,9 +211,13 @@ fn profile_for_delivery(caps: &crate::devcaps::Caps, delivery: TranscodeDelivery
...
         }
     };
+    // Image + text subs the demuxer client-renders (`ff.rs` / the track menu). Leaving a
+    // selected bitmap off made MDE answer transcode, which then 503'd the Original part GET
+    // ("decision is for a transcode"). [`DP_SUBTITLE_CODECS`] is the one list — MDE's
+    // `subtitleStreamID` gate reads the same constant.
     format!(
         "add-direct-play-profile(type=videoProfile&container=mkv,mp4&videoCodec={dp_video}\
-         &audioCodec={dp_audio}&subtitleCodec=srt,subrip,ass,ssa)\
+         &audioCodec={dp_audio}&subtitleCodec={DP_SUBTITLE_CODECS})\
```
GLinnik21: this subtitleCodec list now goes out on every decision incl. HLS, so a selected PGS that used to get burned on a transcode may now be "we render it" and never show up — did you check a PGS + forced transcode?

rust-modules/src/plex/transcoder.rs
```diff
@@ -341,7 +376,9 @@ impl Client {
             .int("directStreamAudio", 1)
             .int("mediaBufferSize", 20971)
             .str("session", session)
-            .str("X-Plex-Session-Identifier", session);
+            .str("X-Plex-Session-Identifier", session)
+            .opt_int("audioStreamID", audio_stream_id)
```
GLinnik21: fwiw on our PMS 1.43.3 the pre-PR build didn't 503 at all (Part GET 200, no /decision), and this branch gets mde=1000 directplay; what's different on your server/account — shared user, or a direct-play restriction?

rust-modules/src/plex/transcoder.rs
```diff
@@ -1050,6 +1123,25 @@ mod tests {
...
+        assert!(super::is_dp_subtitle("MOV_TEXT"));
+        assert!(super::is_dp_subtitle("dvd_subtitle"));
+        assert!(super::is_dp_subtitle("hdmv_pgs_subtitle"));
+        assert!(
+            !super::is_dp_subtitle("vplayer"),
+            "obscure aliases stay off the advertised set"
+        );
+    }
+
     /// PIN: the assumed (table-unreadable) profile is byte-identical to the constant string the
```
GLinnik21: nit: "byte-identical to the constant string before devcaps" isn't true anymore, the literal just grew the subtitle list

rust-modules/src/metadata.rs
```diff
@@ -1612,18 +1612,18 @@ pub(crate) struct PlayingItem {
     pub(crate) subs: Vec<Stream>,
     pub(crate) video_fps: f64, // the played leaf's video fps (0 = unknown) — feeds the Load esInfo
     /// The source's stored frame size, `Media[0]` (0 = unknown). `route.rs`'s local direct-play
-    /// gate tests it against the device table's bound: the smart-DP branch never asks PMS, so
-    /// the profile's `*`-scoped width/height limitation cannot stop a 4K source from reaching a
-    /// 1080p-bounded decoder unless the client also checks it here (issue #22's over-claim
-    /// class — docs/plex-pass-audit.md, closing section).
+    /// gate tests it against the device table's bound: when `/decision` is unreachable the
```
GLinnik21: "the fallback still never asks PMS" reads wrong now, the remux fallback does ask via transcode_decision

rust-modules/src/route.rs
```diff
@@ -5044,6 +5037,36 @@ fn measure_remote_original(
     })
 }
 
+/// PMS 1.43 503s a Part GET after a transcode MDE. The Original we would actually play is a
+/// codec-copy remux, so the Remote capacity sample has to be that `start.mkv`, under the same
+/// session identity a direct Part probe uses.
+fn measure_remote_remux(
+    client: &crate::plex::Client,
+    rk: &str,
+    session: &str,
```
GLinnik21: the remux probe now pays transcoder spin-up under the same 4s budget the raw-Part probe had, so a slow spin-up reads as a slow link and demotes Auto to HLS; also nothing stops the ProgressiveMkv job if the probe bails

rust-modules/src/route.rs
```diff
@@ -5894,7 +5956,24 @@ fn build_stream(rk: &str, part: &str, vcodec: &str, acodec: &str, env: &ResolveE
...
+                            &client.direct_play_url(part, &session).to_url(),
+                            source_transport_kbps,
+                        )
+                    } else {
+                        // Part GET 503s after a transcode MDE. Sample the remux we would actually play.
+                        measure_remote_remux(
+                            client,
+                            rk,
+                            &session,
+                            env.audio_sid,
```
GLinnik21: MDE was asked with audio_sel's id + mde_subtitle_stream_id, but the probe here passes env.audio_sid/sub_sid pre-put_selection — deliberate?

rust-modules/src/route.rs
```diff
@@ -5764,31 +5821,36 @@ fn build_stream(rk: &str, part: &str, vcodec: &str, acodec: &str, env: &ResolveE
...
+        // force a burn.
+        let audio_id = audio_sel.as_ref().map(|(_, _, id)| *id).unwrap_or(0);
+        let subtitle_id = plan
+            .playing
+            .as_ref()
+            .map(|p| mde_subtitle_stream_id(&p.subs))
+            .unwrap_or(0);
+        server_decision(client, rk, &session, audio_id, subtitle_id)
     };
+    let mut directplay = mde.as_ref().is_some_and(|v| v.original);
```
GLinnik21: A timeout or a garbled body lands here too, and then everything transcodes. Maybe keep the old heuristic when the call itself fails, and only fail closed when PMS actually says no?

rust-modules/src/plex/transcoder.rs
GLinnik21, line 220: This goes out on HLS decisions too now. Does a selected PGS still get burned on a transcode?

rust-modules/src/plex/transcoder.rs
GLinnik21, line 385: Couldn't reproduce the 503 on my 1.43.3, the old build direct-plays this shape without any /decision. What's different on your side — shared user, some restriction?

rust-modules/src/plex/transcoder.rs
GLinnik21: Not byte-identical anymore.

rust-modules/src/plex/models.rs
```diff
@@ -406,6 +406,18 @@ pub struct MediaPart {
...
 
+impl MediaPart {
+    /// `/decision` only. `Part.decision=transcode` means the *container* changes; a codec-copy
+    /// remux is still right unless the VIDEO stream's own decision is `transcode` (bit depth,
+    /// an undecodable codec, …). `docs/pms-api.md` measured Profile 5 as `Part=transcode` with
+    /// video `copy`. Absent streams, `copy`, or any other spelling do not forbid a remux.
+    pub fn video_forbids_copy(&self) -> bool {
+        self.stream
+            .iter()
+            .any(|s| s.stream_type == 1 && s.decision == "transcode")
```
GLinnik21, line 419: No Stream[] means copy allowed. Worth a comment on why that's the safe default.

rust-modules/src/metadata.rs
```diff
@@ -1612,18 +1612,18 @@ pub(crate) struct PlayingItem {
...
+    /// fallback still never asks PMS, so the profile's `*`-scoped width/height limitation cannot
```
GLinnik21: The remux fallback does ask PMS now.

tools/plxnative-lab
```diff
@@ -214,6 +214,11 @@ class Upnp:
                     loc = m.group(1).decode()
                     if loc not in locations:
                         locations.append(loc)
+        except OSError:
```
GLinnik21: This swallows every socket error, not just the no-route one.

rust-modules/src/route/decision.rs
```diff
@@ -5044,6 +5037,36 @@ fn measure_remote_original(
...
+        session,
+        true,
+        false,
+        crate::plex::TranscodeOffset::Fresh,
+        audio_stream_id,
+        subtitle_stream_id,
+        None,
+        crate::plex::TranscodeDelivery::ProgressiveMkv,
+    );
+    if client.transcode_decision(&spec).is_none() {
```
GLinnik21, line 4898: The probe now counts transcoder spin-up inside the 4s budget, so a slow spin-up reads as a slow link. Also, who stops this start.mkv job if we bail?

rust-modules/src/route.rs
GLinnik21: MDE got audio_sel's id, the probe gets env.audio_sid. On purpose?

docs/plex-api-catalog.md
```diff
@@ -293,8 +293,21 @@ and `Directory[].key` — **confirmed** (Directory items are `librarySection`, n
...
+  the decision always carries `subtitleStreamID` (an advertised embedded client-rendered
+  id, or `0` so a selected sidecar or unadvertised codec does not force a burn). If
+  `/decision` is unreachable the app does **not**
+  return the Part URL (that would 503); it remuxes or re-encodes instead. An explicit MDE
+  `transcode` still remuxes when the video stream's own decision is `copy` or unnamed
+  (`Part.decision` is a container change); only a video-stream `transcode` forbids the
+  codec-copy. Remote Auto's Original probe then samples that remux `start.mkv` rather than
+  the Part (a Part GET after a transcode decision is 503). (Catalog **D-7** still describes
+  `audioStreamID`/`subtitleStreamID` on *transcode* start/decision GETs as undocumented relative
+  to the PUT selection API; the MDE handshake above is a separate use of those query names.)
```
GLinnik21: D-7 above still calls these no-ops. One of the two is wrong.

GLinnik21 (changes requested):
Tested on my TV against PMS 1.43.3: direct play, many-audio and transcode cases all pass on this branch. Couldn't reproduce the original 503 though — the old build plays the same TrueHD+AC3 item fine, no /decision at all. A few questions inline.

MrcRjs:
> Tested on my TV against PMS 1.43.3: direct play, many-audio and transcode cases all pass on this branch. Couldn't reproduce the original 503 though — the old build plays the same TrueHD+AC3 item fine, no /decision at all. A few questions inline.

[GLinnik21](https://github.com/GLinnik21) TrueHD+AC3 on LAN is the miss, not a contradiction. The old build is supposed to play that without `/decision`. PMS infers local direct play on its own. The 503 is a different item on a **remote** link.

Repro on my side (PMS 1.43.3, Plex Pass, owned server, not a shared user):

1. Client origin has to be **remote**, not LAN. Diagnostics Connection line must read `remote · PMS 1.43.3 · Plex Pass`. If it says `LAN`, you will not see this. Same-LAN TV vs PMS never hits it.
2. Item is **not** TrueHD+AC3. It is 4K DoVi/HDR10, HEVC Main 10, Dolby Vision Profile 8.1, `3840×2072` @ 24 fps, audio **EAC3 5.1 + Atmos**.
3. Open it, then force **Manual → Original** from the player `…` quality menu (`Action: fixed by user`, adaptive controller inactive).
4. On **main / the old build**, the Part GET comes back 503. Diagnostics: `PMS refused Original source · HTTP 503`. PMS log at the same moment:


```
WARN - [Req#bbed] Denying access due to session lacking permission to direct play part 36013
```

That is the 1.43 "no MDE on this session, so the Part GET is refused" path this PR is for. Not a codec veto and not a sharing restriction.

On **this branch**, the same title on the same remote origin registers `/decision?hasMDE=1&directPlay=1` first, then the Part GET is HTTP 200. Sim screenshot attached: Connection `remote`, route `progressive · direct play`, `hevc → H265` / DV P8.1, `eac3 → AC3 PLUS` / Atmos, Manual Original, `HTTP 200 · 10.8 MB received`. Ignore `Load: REFUSED` / "television media pipeline rejected the stream" on that shot. That is the simulator (no Starfish). The 503 is the old-build remote Original Part GET, confirmed by the PMS line above.

image


If you want it on the set: take the TV off that PMS LAN (or force a public/`plex.direct` origin so Diagnostics says `remote`), pick a 4K DV P8 + EAC3 Atmos file, force Original on the old build, grep PMS for `lacking permission to direct play`. TrueHD+AC3 on LAN will keep looking fine.

rust-modules/src/route.rs
MrcRjs: A timeout or a garbled body never registered the session. Restoring the old `unwrap_or_else(is_dp_audio)` heuristic would Part GET anyway, which is the 1.43 503 this PR is for (`Denying access due to session lacking permission to direct play`).

`None` here is not "PMS said transcode". It is "we have no admitted decision". Distinguishing those two does not help Original: only a parsed body with `Part.decision=directplay` is a registered Original, and only that is safe to GET.
This is also not a full reencode. `mde_forbids_copy` is false on `None`, so a video DP item still remuxes (`start.mkv` copy) by a separate `transcode_decision`. What we must not do is hand back the Part URL.
Leaving this as fail closed.

rust-modules/src/plex/transcoder.rs
MrcRjs: Yes. The new `subtitleCodec=` list is on `add-direct-play-profile`, which is what MDE uses for Original. HLS / reencode is a different surface: `transcode_query` still sends `directPlay=0` and, when `subtitle_stream_id > 0`, `subtitles=burn` + `subtitleStreamID`. The HLS target is still `mpegts` / h264 / aac, never a PGS copy.

Advertising PGS on the DP profile is what stops MDE from answering `transcode` for a selected bitmap we already client render. It does not opt HLS into soft PGS. Soft subs still do not deliver on this pipelin, transcode still burns.

No change here unless a live HLS `/decision` body shows a selected PGS as `copy` which I haven't seen before.

rust-modules/src/plex/transcoder.rs
MrcRjs: I left another comment with more details for repro https://github.com/GLinnik21/plx-native/pull/78#issuecomment-5627269748

rust-modules/src/plex/transcoder.rs
MrcRjs: Agreed. I’ll retitle / recomment it to: assumed caps still interpolate `DP_SUBTITLE_CODECS` and any further profile edit must update this literal. The assertion itself stays.

rust-modules/src/metadata.rs
MrcRjs: Yes. The local width/height gate is still the skip before MDE (`!video_dp` → `skip_mde`), so a 4K source on a 1080p-bounded SoC never relies on the profile limitation being honoured by a missing `/decision`.

What went stale is “the fallback still never asks PMS”. Unreachable MDE now remuxes through `transcode_decision` instead of returning the Part. I’ll fix that sentence.

rust-modules/src/route/decision.rs
MrcRjs: Two separate things.

Spin up inside the 4s window is real. A remux `start.mkv` is not a Part GET. Encoder start looks like a slow link. For Remote Auto that bias is toward HLS, which is the safe direction. I am not going to pretend the first byte wait is transport capacity. A follow up can exclude time to first byte if a slow NAS keeps losing Original on a fast WAN.

Stop on bail is missing. If the sample fails and we go HLS, this leaves a remux job on the playback session. I’ll `transcode_stop` when the probe returns `None`. If the sample succeeds and we actually play that remux that job is the playback so it must not be stopped here.

rust-modules/src/route.rs
MrcRjs: Not on purpose. MDE names the smart-DP sibling (`audio_sel`’s AC3/EAC3/AAC id) so TrueHD/DTS does not veto Original. `env.audio_sid` is `cur_audio_sid` at resolve start (often `0` on a cold play) which is the part default.

On a TrueHD+AC3 item that remuxes, the probe would sample a TrueHD remux while `/decision` was for the AC3 sibling. Copilot's preflight note is the same split.

I'll pass the same `audio_id` / `subtitle_id` already used for `server_decision`.

docs/plex-api-catalog.md
MrcRjs: Agreed. One of the two is wrong, and it is D-7.

On **MDE** (`hasMDE=1`, `directPlay=1`) `audioStreamID` / `subtitleStreamID` are load bearing. Omitting the chosen AAC/AC3/EAC3 sibling is what made MDE evaluate TrueHD and 503 the Part GET. `subtitleStreamID=0` is what stops a selected sidecar from forcing a burn.

On **transcode** start/decision (`directPlay=0`) the PUT `/library/parts/{id}` is still the selection API. The GET copies remain an extra safe match to that PUT.

I’ll rewrite D-7 to that split and drop the “safe to drop” line for MDE.

rust-modules/src/plex/transcoder.rs
```diff
@@ -341,7 +376,9 @@ impl Client {
             .int("directStreamAudio", 1)
             .int("mediaBufferSize", 20971)
             .str("session", session)
-            .str("X-Plex-Session-Identifier", session);
+            .str("X-Plex-Session-Identifier", session)
+            .opt_int("audioStreamID", audio_stream_id)
+            .int("subtitleStreamID", subtitle_stream_id);
```
GLinnik21: This needs an explicit subtitle mode too. With an embedded SRT already selected, I get HTTP 400 on the TV. PMS says:

```
MDE: hasMDE was specified along with directPlay on a part with a selected subtitle and invalid subtitle setting 'auto'
```

Then we fall back to remux and PUT `sub=0`, so the selected subtitle disappears. Can we set the client-rendered mode here and cover a subtitle selected before Play? The current tests check the id, but miss this rejection.

rust-modules/src/route.rs
```diff
@@ -5888,13 +5960,35 @@ fn build_stream(rk: &str, part: &str, vcodec: &str, acodec: &str, env: &ResolveE
...
+                        )
+                    } else {
+                        // Part GET 503s after a transcode MDE. Sample the remux we would actually play.
+                        remux_probed = true;
+                        measure_remote_remux(
+                            client,
+                            rk,
+                            &session,
+                            audio_id,
+                            subtitle_id,
```
GLinnik21: The probe gets the sibling now, but the final `put_selection` and `transcode_spec` below still use `env.audio_sid`. I extended `remote_auto_truehd_remux_probe_names_the_ac3_sibling_not_env_audio_sid` to check the final URL too: the probe asks for AC3 id 2, playback asks for TrueHD id 1. Can we carry the same selection through both? This one is from the host test; I haven't exercised Remote Auto on the TV.

GLinnik21 (commented):
Retested this head on my TV against PMS 1.43.4: 7 passed, 1 failed. TrueHD+AC3 and DV P8 direct-play fine. Found a selected-subtitle regression on LAN this time; PMS logs explain the 400. Also checked the audio-id change — the probe is fixed, but the final playback request still differs. Details inline.

rust-modules/src/plex/transcoder.rs
MrcRjs: Agreed. Naming the stream id without a mode leaves PMS on `auto`, which 1.43.4 400s `hasMDE`+`directPlay` when a subtitle is already selected. I'll send the client-rendered mode on the MDE query and add a test that would have caught this, so we stay Original instead of remux + PUT `sub=0`.

MrcRjs: Agreed. Naming the stream id without a mode leaves PMS on auto, which 1.43.4 400s hasMDE+directPlay when a subtitle is already selected. I’ll send the client-rendered mode on the MDE query and add a test that would have caught this, so we stay Original instead of remux + PUT sub=0.

rust-modules/src/route.rs
MrcRjs: The probe names the AC3 sibling now; put_selection and transcode_spec still use env.audio_sid. I’ll carry the same audio_id through both and extend that host test to assert the final URL.

rust-modules/src/route/plan.rs
```diff
@@ -1219,15 +1242,25 @@ pub(super) fn build_stream(rk: &str, part: &str, vcodec: &str, acodec: &str, env
     // reasoning `remux` and `no_video_copy` carry: a seek and an audio switch rebuild this query
     // from `Session`, and one that dropped the ceiling would hand the encoder back the full
     // 4K/60 Mbps bound the moment the user touched the scrubber.
-    put_selection(env.sid, plan.part_id, env.audio_sid, env.sub_sid); // audio/subtitle selection drives the encode/remux + burn
+    // Same audio id MDE and the remux probe already named. `env.audio_sid` is the part default
+    // (TrueHD) at resolve start; putting that undoes smart-DP. Subtitle stays `env.sub_sid`:
+    // a positive id here is a burn, and Original client-renders instead.
+    put_selection(env.sid, plan.part_id, audio_id, env.sub_sid);
```
GLinnik21: This changes the re-encode path too. With English DTS selected, starting at 720p PUTs the Russian AC3 sibling instead; the TV played it and PMS confirmed the source track. Can we preserve the selected audio when re-encoding?

GLinnik21 (commented):
Retested this head on my TV with PMS 1.43.4. The selected SRT now stays on Direct Play, thanks. The audio change still has a problem on the re-encode path — reproduced on the TV this time, details inline.

rust-modules/src/route/plan.rs
MrcRjs: Sure. The AC3 sibling is for remux so we can copy something the TV can play. A 720p reencode can transcode the selected DTS, so I'll split those ids in a follow up on this PR and keep the chosen track.

GLinnik21 (commented):
Retested on my TV with PMS 1.43.4: 720p now keeps the selected English DTS, confirmed in the PMS source mapping. Selected SRT and TrueHD+AC3 Direct Play still pass too. That clears the regressions I found, thanks.

GLinnik21 (approved):

PR #79, opened 2026-09-11
Settings: Automatically Sign In skips the profile picker on boot

MrcRjs:
On a multiuser Plex Home, every interactive boot still asked who's watching. Solo accounts already skipped that screen. This adds the opt in toggle to automatically select a profile.

Settings -> System gets an **Automatically Sign In** switch when you're signed in and the roster has more than one profile. It sticks on the session file as `auto_sign_in` (missing or junk counts as off, which is the picker everyone already knows), written through `session::update` so a roster or recents write can't quietly drop it. A full profile-switch `save` keeps it too.

Boot goes through `Session::boot_shows_picker`. The switch only skips the picker when the seated uuid is still on the roster. Empty uuid (walked away at the picker) and a uuid that isn't in `home_users` any more (removed user, stale roster) still get the list. Automated boots and `plxnative-pickuser` behave as before. 

Note: Turning this on while a PIN protected profile is active means the next launch skips that PIN on purpose.

Change Profile -> BACK still won't resume for you.

`docs/agent-reference.md` and `docs/plex-pass-audit.md` mention the new gate.

Host tests cover the soft-parse, the persist/`save` keep, the picker table(including `u-gone`), and the Settings row (toggle in place, RIGHT doesn't push, signed-out hides it). On a set: turn it on under a PIN profile, relaunch straight into Home. turn it off and the picker comes back. Change Profile, relaunch, you land as whoever you just picked.

https://github.com/user-attachments/assets/6c0be010-6b73-4ad6-8bc3-9313585e4dd8

### DIFF
--- a/docs/agent-reference.md
+++ b/docs/agent-reference.md
@@ -1644,7 +1644,7 @@ path. Never run only this one before a release. `tests/README.md` has the tier t
   `tests/run.py` always injects (it reads the owner token from the gitignored
   `src/config.local.h` on the HOST; that macro is never compiled in). An interactive boot with
   no session lands on the QR sign-in screen.
-- Normal interactive flow: who's-watching picker (multi-user) → Home; D-pad/pointer to focus a
+- Normal interactive flow: who's-watching picker (multi-user, unless Automatically Sign In is on) → Home; D-pad/pointer to focus a
   card → **OK** opens the detail page → Play starts playback; OK toggles play/pause, LEFT/RIGHT
   scrub-seek, **BACK/Stop** returns. The strip's **last pill is Search** (a mark, not a word) — a
   peer of Home and the Library, not a page stacked over them, so BACK from it returns to Home. BACK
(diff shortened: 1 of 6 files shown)


GLinnik21:
I was thinking exactly the same thing, and I believe the original Plex app has a setting like this too. I’m still not entirely sure about the expected behavior, though:

1. If the currently selected profile is PIN-protected, does enabling automatic sign-in mean the PIN will be skipped on future launches?
2. With multiple users, does automatic sign-in always use the last profile that signed in?

If so, that behavior makes sense to me — I just want to confirm.

MrcRjs:
> I was thinking exactly the same thing, and I believe the original Plex app has a setting like this too. I’m still not entirely sure about the expected behavior, though:
> 
>     1. If the currently selected profile is PIN-protected, does enabling automatic sign-in mean the PIN will be skipped on future launches?
> 
>     2. With multiple users, does automatic sign-in always use the last profile that signed in?
> 
> 
> If so, that behavior makes sense to me — I just want to confirm.


1. Yes. If you turn this on while a PIN protected profile is already signed in, the next launch skips the PIN and goes straight in. It only applies at boot. Change Profile still asks for a PIN and if the stored profile is gone from the household, or nobody was actually seated last time, you get the picker again.

2. Yes, last person who sat down. It is not tied to whoever flipped the switch. If Kid turns it on and Admin uses the TV later, the next launch is Admin.

GLinnik21 (approved):
Approved after real-TV verification on the unchanged PR head. Automatically Sign In appeared for the multi-profile account, persisted On across a clean interactive relaunch into Home for the PIN-protected active profile, and Off restored the profile picker. Host, no-default-features, and ARM build gates passed.

PR #80, opened 2026-09-11
ui: a thicker white rim on the focused chapter so it reads over the video plane

MrcRjs:
The Chapters strip sits on the hardware video plane. The card family's resting sheen (`theme::CARD_SHEEN`, .22 / 1px) washes out there the same way an unkeyed control's edge did, so LEFT/RIGHT through the strip did not say which still was selected from the couch.

## What this PR changes

The focused chapter now draws a round-rect rim in `theme::CONTROL_RIM_FOCUS_UNKEYED` (the unkeyed control's pure-white focus edge) at **2.5px**. That width is a step over the control family's 1.25: a 288-wide still needs more than an edge to read as selected, not merely edged.

The rim is painted on the same scaled frame `card()` draws into, so it rides the focus pop instead of lagging a resting box. Full strength whenever focused: the pop already animates the geometry, and fading the rim with it would blank the selection mark on every LEFT/RIGHT.

`CH_RAD` and `CH_FOCUS_RING_W` name the numbers that were previously a bare `10.0` into `draw_card`.

## Verification

Layout and colour over a moving picture. `make check` cannot see a pixel. The simulator cannot see the video plane, which is the whole reason the sheen failed.

- `make check`
- Simulator: open Chapters during playback, walk LEFT/RIGHT, confirm the rim rides the pop and does not vanish mid-move

### Before

image

image

### After

image

image

### DIFF
--- a/rust-modules/src/ui/chapters_panel.rs
+++ b/rust-modules/src/ui/chapters_panel.rs
@@ -19,6 +19,13 @@ const CH_W: f32 = 288.0;
 const CH_H: f32 = 162.0; // 16:9 still
 const CH_GAP: f32 = 24.0;
 const CH_TOP: f32 = 684.0; // thumbnail top — name/time fit above the tabs (SCR_H-128)
+const CH_RAD: f32 = 10.0;
+/// Focus rim on the selected chapter. The card family's resting sheen ([`theme::CARD_SHEEN`] .22 /
+/// 1px) washes out over the hardware video plane the same way an unkeyed control's edge did — pure
+/// white and a thicker stroke are what keep "which chapter" readable from the couch while scrolling.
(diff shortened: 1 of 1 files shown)


GLinnik21:
Cool, thanks! I was thinking the same. I may have already removed the white border shader, but I need to run the FPS suite on my old TV before I can say for sure. That hardware is pretty fragile performance-wise, and even a small shader can make a noticeable difference, so I’d rather test it first.

I also have a broader goal of improving the focused state of all tiles to make the currently focused item easier to spot — just not with a white border :) But for now, I think this one is fine as is.

GLinnik21 (approved):
Approved after real-TV verification on the unchanged PR head. The focused chapter rim remained clearly readable over live HDR video and followed focus across multiple chapters. Profiler-free production pacing held 60.0 FPS median (p10 59.2, p95 60.0; n=13). Host, no-default-features, and ARM build gates passed.

PR #89, opened 2026-09-13
Sequential media GETs reuse the HTTP/1.1 demux socket

MrcRjs:
Media socket is closing after every sequential GET. Each HLS segment and each Original range paid a new TCP handshake for a server that already speaks HTTP/1.1 keep-alive. That shows up as a stall, not as tidy cleanup.

## What this PR changes

Media requests now leave `Connection` off so PMS can keep the fd. The next GET to the same peer reuses it once the previous body is fully read. An unread body, leftover buffer bytes, HTTP/1.0, or an explicit `Connection: close` still opens a new socket.

If the peer idles out or half closes, we redial once instead of failing the GET. A shutdown of a live keep-alive fd is teardown of that open. We do not dial a replacement the shutdown cannot reach.

Chunked responses consume trailers before reuse, so leftover bytes are not mistaken for the next status line. Incomplete trailers close the socket.

`crate::http` is unchanged: it still sends `Connection: close` and still closes a one shot stream. HTTPS media reuses the libcurl multi across playlist and segment URLs. A Range seek still closes first, so tests can still tell a seek from a reused GET by counting accepts.

Keep-alive only helps if the socket stays readable. Original parks `aq_push` on the AVIO thread, and a full AU queue used to let the TCP window collapse (or let PMS FIN us). While parked we now drain available bytes into the bounce buffer. That is why reuse survives a backlog.

Docs that still said media used `Connection: close` are updated. Main-thread control-plane freezes are the same as before.

## Verification

Host tests in `stream.rs`, `aq.rs`, `http.rs`, `curlio.rs`, and `ff.rs` cover:

- two drained HTTP/1.1 and chunked GETs share one accept
- `Connection: close`, an unread body, HTTP/1.0 keep-alive, and incomplete trailers do not reuse
- named chunked trailers reuse; poisoned leftovers do not
- media sequential GETs omit `Connection: close`; control-plane plaintext still sends it
- shutdown of a live keep-alive fd does not dial a replacement
- idle peer close redials the next GET; teardown after that close does not
- drain returns immediately when the body is done or the peer stalls mid-chunk
- mid-body EOF is a transport error on both read and drain
- `aq_push_with_drain` runs while the consumer holds the queue full
- curl sequential GETs reuse the multi for a second GET

- `make check`

Reuse and drain are host-visible. Whether Original holds a live socket across a real AU backlog still needs a TV play.


rust-modules/src/stream.rs
```diff
@@ -922,15 +1019,65 @@ fn http_open_with_timeouts(
...
+                Err(HttpOpenError::Aborted) => return Err(HttpOpenError::Aborted),
+                Err(HttpOpenError::Deadline) => return Err(HttpOpenError::Deadline),
+                Err(HttpOpenError::Status(status)) => return Err(HttpOpenError::Status(status)),
+                Err(HttpOpenError::Transport) => {
+                    // Idle timeout or a half-closed keep-alive: one redial, not a hard failure.
+                    close_owned(hs);
+                    if hs.interrupted() {
+                        return Err(HttpOpenError::Aborted);
+                    }
+                    hs.reset_fields();
```
GLinnik21: There is a race here: a shutdown between the first check and reset_fields() gets lost, so teardown can still open a new socket; preserve the latch across the reset.

rust-modules/src/stream.rs
```diff
@@ -1329,6 +1558,258 @@ pub(crate) fn http_close(hs: *mut HttpStream) {
...
+        if hs.fd() < 0 && (hs.bpos as usize) >= (hs.blen as usize) {
+            return hs_closed_or_idle_read_result(hs);
+        }
+        if hs.content_length >= 0 && hs.consumed >= hs.content_length {
+            finish_body(hs);
+            return 0;
+        }
+        if hs.chunked != 0 && hs.chunk_left < 0 {
+            let due = Instant::now();
+            match hs_skip_chunked_trailers(hs, Some(due)) {
```
GLinnik21: If a chunked trailer arrives in two TCP fragments, deadline=now leaves body_done false forever because the next drain never reads from the socket.

rust-modules/src/stream.rs
```diff
@@ -2632,4 +3115,798 @@ mod tests {
...
+        std::thread::scope(|sc| {
+            sc.spawn(|| {
+                let Ok((mut s, _)) = srv.accept() else { return };
+                let mut buf = [0u8; 2048];
+                let n = s.read(&mut buf).unwrap_or(0);
+                assert!(
+                    !buf[..n]
+                        .windows(b"Connection: close".len())
+                        .any(|w| w.eq_ignore_ascii_case(b"connection: close")),
+                    "media sequential GETs must omit Connection: close so the fd can reuse"
```
GLinnik21, line 3493: The unit test covers this, but tests/serve_fixtures.py still replies with Connection: close, so the local pipeline never exercises connection reuse.

GLinnik21 (changes requested):
There are a couple of nasty edge cases here, and the local pipeline still does not exercise connection reuse. Please fix these before merging.

MrcRjs:
> There are a couple of nasty edge cases here, and the local pipeline still does not exercise connection reuse. Please fix these before merging.

@GLinnik21  Agree. The live-fd reuse path already captures interrupted across reset_fields. The redial-after-Transport path does not, drain can stall on split chunked trailers, and tests/serve_fixtures.py still answers Connection: close, so the pipeline never reuses. Will fix those three and push.

rust-modules/src/stream.rs
MrcRjs: Fixed. The live-fd path already kept `interrupted` set on abort. The Transport redial path called `reset_fields()`, which cleared the latch, so a teardown that landed after the reuse send looked like a later session and dialled a socket the shutdown cannot reach.

reset_request_fields() no longer touches the latch. A leftover interrupt is consumed only when fd < 0 at the start of http_open (the later-session contract). Transport redial never consumes it. teardown_during_keepalive_transport_redial_does_not_dial pins accepts stay at 1.

rust-modules/src/stream.rs
MrcRjs: Fixed. recv_until returns HTTP_READ_DEADLINE before poll/recv when the deadline is already due, so drain's Instant::now() never read the socket. Chunk-size lines already filled with hs_fill_buf_dontwait first; trailer skip did not.

Both trailer sites in drain now go through hs_finish_chunked_trailers_dontwait: one dontwait fill, then parse with deadline=now against buf only. Payload drain is unchanged. chunked_drain_completes_trailers_split_across_two_writes covers the two-fragment case and still reuses one accept.

rust-modules/src/stream.rs
MrcRjs: Fixed. The unit test only proved the client omits Connection: close. tests/serve_fixtures.py still sent it on every 2xx and set close_connection = True, so can_reuse never saw keep_alive.

2xx playlist and media replies now omit that header and leave the HTTP/1.1 handler looping. _fail still closes. Spec item 1 matches stream.rs. FixtureKeepAlive in tests/test_harness.py pins two sequential GETs on one accept, Range 206 on the kept-alive fd, and a 404 that still closes. stats() is still (n_opens, n_ranged).

rust-modules/src/stream.rs
GLinnik21: This is still only partially fixed: splitting after `Foo: bar` and before `\r\n\r\n` makes the new test fail because `line_empty` is reset between parser calls; please persist the trailer parse state and cover that boundary.

rust-modules/src/stream.rs
```diff
@@ -922,15 +1018,65 @@ fn http_open_with_timeouts(
...
+        // A shutdown of a live keep-alive fd is teardown of THIS open. Closing it and returning
+        // leaves `interrupted` set so we do not dial a replacement the already-fired shutdown
+        // cannot reach. `fd < 0` with a leftover interrupt is a later session: consume it here
+        // only — that is the contract `a_new_open_clears_the_interrupt_from_the_last_one` pins.
+        if hs.interrupted() {
+            if hs.fd() >= 0 {
+                close_owned(hs);
+                return Err(HttpOpenError::Aborted);
+            }
+            hs.interrupted.store(0, Ordering::Release);
```
GLinnik21: This can still clear a live teardown: shutdown may land after the caller’s AU-abort check while fd is -1, then this store erases the latch and the open connects under join; a fresh HttpStream per engine means there is no old production session latch to consume here.

GLinnik21 (commented):
The retry race is fixed, but the broader entry race is still here.

GLinnik21 (commented):
The real TV run exposed one fixture-server regression from enabling keep-alive.

tests/serve_fixtures.py
```diff
@@ -130,6 +131,10 @@ class FixtureHandler(BaseHTTPRequestHandler):
     def log_message(self, fmt, *args):  # noqa: A003  (BaseHTTPRequestHandler's name)
         self.server.note(f"{self.address_string()} {fmt % args}")
 
+    def setup(self):
```
GLinnik21, line 134: A normal TV seek or teardown resets this kept-alive socket while BaseHTTPRequestHandler is waiting for the next request, so every case now prints an uncaught ConnectionResetError; catch the reset around the persistent request loop too.

GLinnik21:
For device verification, you can ask an agent to run the universal synthetic player tier (`./tests/run.py`). It serves generated dummy media and covers Original, HLS, Auto/ABR, network shaping, seek, EOS, and other playback paths without needing a Plex library. The real PMS cases under `--server` were written around my private library because I did not expect external contributors, so I do not expect you to reproduce those. If you can think of a better portable way to verify this end to end, feel free to propose it.

MrcRjs:
> For device verification, you can ask an agent to run the universal synthetic player tier (`./tests/run.py`). It serves generated dummy media and covers Original, HLS, Auto/ABR, network shaping, seek, EOS, and other playback paths without needing a Plex library. The real PMS cases under `--server` were written around my private library because I did not expect external contributors, so I do not expect you to reproduce those. If you can think of a better portable way to verify this end to end, feel free to propose it.

Thanks! Yeah I've just doing some manual testing along with running some partial tests. Long term we can probably looks at something like https://github.com/iwalton3/stdjflib or our plex-variation to run a full regression and programmatically generate a PMS test library.

GLinnik21:
> programmatically generate a PMS test library.

As you can see, I'm doing a huge refactoring of the UI architecture. Turns out GPT Astra generates such mocks on the fly 😅 to check behaviour

GLinnik21 (dismissed):
Update: I ran the same two cases against current `main` (`6073ae00cfe17e3cea58d19e50e734f4fe9cc2c5`) on the same TV, network, and fixture pack. Both fail there too, with essentially the same numbers:

- `pipe_replay_after_eos`: again 0 `replay:` lines and one body instead of two.
- `pipe_abr_down_collapse`: again only `pos_climb` fails — 37s instead of the required 45s (the PR run reached 36s).

So these failures were not introduced by #89 and should not block it. The PR run itself finished 42/44; all keep-alive, HLS/ABR wire, and Range-reuse checks passed. Sorry for the noise — the baseline control clears the PR.

GLinnik21 (approved):

PR #92, opened 2026-09-13
Add trailer playback from movie and show detail

MrcRjs:
Official movie and show pages offer trailers. Ours had Related and Cast and never asked PMS for extras, so there was no disc, no menu row, and no clip to play.

## What this PR changes

Movie and show detail fetch `GET /library/metadata/{rk}/extras` in parallel with `/related`. Serial depth stays 2 / 5; a refused extras GET does not fail the page. Episode and season pages do not pay that hop and do not inherit the show trailer.

The picker keeps **one** playable trailer (`subtype == "trailer"` or `extraType == 1`, with a non-empty `Part.key`), preferring `primaryExtraKey`, else first in server order. A primary that names a non-trailer, or a trailer with no Part, is ignored. At most one follow-up `metadata()` fills a missing Part; never N+1 over the rest of the list.

When that extra is playable, the hero grows a Trailer disc (after Play / Resume, before *Also available*), and the item menu offers Play Trailer. The menu is cache-only off the already-loaded parent: a related tile cannot steal it, and episode / season menus never offer it. No extras means no disc and no row.

Playback reuses `request_play` with the extra's rk / part, `resume_ns = 0`, HUD context `"Trailer"`, and `continuous` omitted so EOS cannot Up-Next into a sibling extra. `up_next_of` is already episode-gated; a clip successor would not arm the tile even if a queue came back larger than 1. `metadata::current()` stays the parent, so Go to Show after a show trailer still opens the show.

Quality judging uses the extra's own bitrate. Using the parent's 4K figure (or `src_kbps = 0`, which `Ceiling::admits` fails closed on) would force every non-Auto rung through the encoder.

The extras **shelf** (behind-the-scenes, featurettes, interviews) is still out of scope. `Detail` keeps `trailer: Option<Extra>`, not the extras vector. Parity gaps, the Plex Pass audit, and `docs/pms-api.md` §4 say so.

Supporting: `DetailScreen` caches hero-chain and below-hero section geometry for the frame. Synopsis height and the episode strip's `block_h` are O(text) and used to be re-asked from every `place` in `record_stops`. The cache is stamped off counts, string lengths, and pointer identity so an in-place season landing (`pump_season` rewriting `CURRENT` at a stable address) still rebuilds.

`SCREEN_SHAPES_PIN` moves because `Action::PlayTrailer` grows the item-menu shape with the extra's play fields.

## CI fix

`hostsim`-feature host checks were failing to compile: `player::engine`'s `load_in_flight_tests` module (and two jail-refusal tests in `player::mod`) still called the pre-restructure API — free functions `engine_take`/`engine_install`/`engine(&mt)`, and `state()`/`start_bufferfeed(&mt)`/`reset_player_control_for_test()` with the old argument-free signatures — left behind by the `PlaybackSession`/`PlayerAdapter` refactor that landed earlier on this branch. Rewrote them to the pattern already used correctly elsewhere in the same file (`load_refusal_tests`, `replay_after_stop_tests`): a local `PlayerAdapter`/`PlaybackSession` threaded through `pump()`/`start_bufferfeed()`, `pa.engine()`/`pa.install()` instead of the retired globals. Default-feature and `hostsim`-feature builds both now compile clean under `cargo test --lib --no-run`.

## Verification

Host tests cover the picker (primary preference, non-trailer primary ignored, extraType 1, empty Part), extras-wanted only for movie / show, the disc playing the extra from the start, a show disc not playing the on-deck episode, extras landing growing the disc without moving Play identity, a movie without extras offering nothing, Play Trailer cache-only on the loaded parent, trailer PlayQueues omitting `continuous`, a clip queue row not arming Up-Next, quality judging the extra file not the parent or zero, and Go to Show after a trailer opening the loaded parent.

Device / simulator still have to see the disc, the HUD context, and that EOS does not Up-Next. `make check` cannot see a pixel, and the extras GET is live PMS.


GLinnik21:
Tested `436d2afe` on the LG webOS 4.5 TV. I found two playback issues that should be addressed before merging.

### 1. Original trailer playback builds a malformed URL

The trailer returned by PMS has a `Part.key` that already contains query parameters, for example:

```text
/services/iva/assets/.../video.mp4?fmt=4&bitrate=5000
```

`Client::direct_play_url()` passes this key to `QueryBuilder`, and `QueryBuilder::build()` unconditionally inserts another `?` before the playback identity parameters. The request sent by the TV therefore has this shape:

```text
/services/iva/assets/.../video.mp4?fmt=4&bitrate=5000?X-Plex-Session-Identifier=...
```

The device log shows the resulting failure:

```text
decision: part=directplay ... -> DIRECT PLAY
stream: GET /services/iva/assets/.../video.mp4 status=400
ff: http_open FAILED status=400
ff: demux produced no access units — treating as a failure
```

With fixed `Original` quality this surfaces as a playback error. In `Auto`, the same Original request fails first, after which the player successfully falls back to HLS.

Relevant code: `rust-modules/src/plex/library.rs::Client::direct_play_url` and `rust-modules/src/plex/client.rs::QueryBuilder::build`.

The query builder should preserve an existing query string and append new parameters with `&`. Please add a regression test using a `part_key` that already contains `?fmt=4&bitrate=5000`.

### 2. Show trailers are labelled `S0 · E0` in the player

`metadata::trailer_now_playing()` currently constructs show-trailer metadata with:

```rust
is_episode: d.is_show,
season: 0,
index: 0,
```

That selects the episodic HUD layout for a show trailer, so the player renders the placeholder values as `S0 · E0`. A trailer is a show-level extra, not episode zero of season zero, and should not display an episode label.

Please add coverage for the player/HUD descriptor produced for a show trailer, asserting that no season/episode label is shown.

GLinnik21 (changes requested):
Thanks for working on this — trailers are something I've wanted to add for a long time too 🙌

I had a slightly different UX in mind, though. Rather than copying the standalone Trailer button from the original Plex app, I was thinking more along the lines of autoplaying the trailer in the Hero area instead of the static Hero image, like most of the big streaming apps do these days.

Another option would be an Extras section on the detail page, where trailers and the other extras are shown as preview tiles. That would make the trailer feel like part of the item's additional content instead of a one-off action.

I'm not strongly attached to one of those two options yet, but I'd like us to settle that direction before locking in the button-only approach. Together with the two playback issues in my comment above, I'm requesting changes for now.

MrcRjs:
> Thanks for working on this — trailers are something I've wanted to add for a long time too 🙌
> 
> I had a slightly different UX in mind, though. Rather than copying the standalone Trailer button from the original Plex app, I was thinking more along the lines of autoplaying the trailer in the Hero area instead of the static Hero image, like most of the big streaming apps do these days.
> 
> Another option would be an Extras section on the detail page, where trailers and the other extras are shown as preview tiles. That would make the trailer feel like part of the item's additional content instead of a one-off action.
> 
> I'm not strongly attached to one of those two options yet, but I'd like us to settle that direction before locking in the button-only approach. Together with the two playback issues in my comment above, I'm requesting changes for now.

That's a good idea, I like that UX that it loads and plays the trailer automatically and then having the extra/alternative trailers in the detailed view. I'll update this PR to address your findings and implement that UI/UX for the trailers.

GLinnik21:
Btw, I have a design system in Claude Design, along with mockups for all the screens, but I don’t think I can share them on a personal plan.

MrcRjs:
@GLinnik21 How about we wrap up the changes here as an initial simplified version, and you can follow up later with the designs you have?

At this point, it gets the trailers and autoplays them in the background from the detailed view. I also added a `Design.md` document with more details on the decisions that were made.

MrcRjs:
I have a few more changes in progress for the UI style for the title logo and the interactions when focusing on the trailer, I will push those changes in a few hours.

GLinnik21 (approved):
Wow, this is super cool! I really appreciate the animations. I have two minor suggestions:

- Perhaps “Extras” should be placed under “Cast & Crew.”
- Maybe the trailer should play only once instead of looping.

docs/DESIGN.md
GLinnik21, line 1: Maybe better move it to `docs`

MrcRjs:
> Wow, this is super cool! I really appreciate the animations. I have two minor suggestions:
> 
>     * Perhaps “Extras” should be placed under “Cast & Crew.”
> 
>     * Maybe the trailer should play only once instead of looping.

Agree, I'll change the order and prevent the looping. What do you think about the play button staying visible even in fullscreen play? Right now it fades out but it's focused when you go fullscreen so if press "enter" it play the media.

GLinnik21:
> Right now it fades out but it's focused when you go fullscreen so if press "enter" it play the media.

Oh, that wasn’t obvious to me. I accidentally pressed Up and only then realized that it hides the UI. So yeah, having Enter activate a hidden button doesn’t seem like good UX. In that state, I’d expect the controls to affect the trailer playback instead.

GLinnik21 (approved):
I think we can merge this now — there’s already a lot of good work here and I’d rather not keep extending the PR.

One thing I’d like to revisit separately: after trying the latest version, keeping the main Play button visible by itself in fullscreen trailer mode looks a bit odd to me. I think the better long-term behavior would be to hide the detail-page actions completely and have Up/Enter show the trailer playback HUD instead.

I’ll track that as a follow-up rather than blocking this PR on it.

PR #102, opened 2026-09-16
fix: pump the async season landing every frame, or the episode spinner never clears

MrcRjs:
Switching season tabs on a show's detail page left the episode row spinning forever and refused every episode press. The season fetch landed on a worker thread, but nothing on the main thread ever read it back.

## What this PR changes

`crate::stores::metadata::pump_season()` drains the season fetch's mailbox and applies it — the episode list, `cur_season`, and the flag `season_loading()` reports, which the episode row's spinner and `episodes::action`'s press gate both key off. Nothing called it. The legacy `ui/detail.rs::update()` used to pump it every frame while the Detail page was mounted; the phase-7 owned-screens migration (894f20f8) deleted that function and never added an equivalent call anywhere — not in the new `screens/detail/mod.rs`, not beside `pump_detail()` / `pump_alt_sources()` in `app/run.rs`'s route-unconditional block, where `stores/metadata.rs`'s own doc comment already claimed it belonged. With no reader, `SEASON_DONE` never caught up to `SEASON_GEN` and `season_loading()` stayed true for the rest of the session after the first season switch.

Added the missing `pump_season()` call next to `pump_detail()` in `app/run.rs`. Also corrected the doc comment in `stores/metadata.rs`: it described these pumps as reached through `MetadataStore::step`'s `StoreEv::Pump` arm, which nothing in production ever sends (only `Bridge`'s owned `BrowseStore` gets a `Pump` event, via `StoreWork::Browse`) — the arm was dead code, and the comment now names the real call sites.

## Verification

`app/run.rs::update()` has no host fixture — it needs a real SDL/GL `App` — so the regression test pins the source text itself, the same idiom `app/words.rs`'s `focusprobe_player_overlay_delegates_to_the_shared_overlay_word_function` already uses for this class of wiring nothing can otherwise exercise: `stores::metadata::pump_wiring_tests` asserts `pump_season()` sits beside `pump_detail()` in `run.rs`. Watched red against the pre-fix source (the call absent) and green with it restored.

- `make check`

The source pin proves the call is wired in; it can't see the picture on screen. A device or simulator pass through an actual season switch is still owed.


ci/allow/sibling-migration.txt
```diff
@@ -1,10 +1,9 @@
-# count: 5
+# count: 0
```
GLinnik21, line 1: heads up, main already carries this exact fix (the count:0 resolution) via 910b947b — this hunk's gonna vanish once you rebase, might as well drop it now so the PR stays about the season bug

ci/check-deps.sh
```diff
@@ -132,6 +132,7 @@ strip_strings_and_comments() {
     }
     print out
   }' "$1"
+}
```
GLinnik21, line 135: same story as the sibling-migration.txt hunk — main already has this closing brace from 910b947b, this'll no-op on rebase

rust-modules/src/app/run.rs
```diff
@@ -2069,6 +2069,18 @@ pub(crate) unsafe fn update(app: &mut App, fr: &mut Frame) {
...
+        // Async season load: install the worker's episode list into CURRENT. Route-unconditional
+        // for the same reason as pump_detail above — this used to be pumped from the legacy
+        // `ui/detail.rs::update()`, which only ran while the Detail page was on top; the phase-7
+        // migration to an owned `DetailScreen` (894f20f8) deleted that call site and never added
+        // an equivalent one, so `pump_season()` was never invoked in production at all. Nothing
+        // drained `SEASON_RESULT`, so `SEASON_DONE` never caught up to `SEASON_GEN`:
+        // `season_loading()` stayed true forever after the first season switch, wedging the
+        // episode row's spinner on and refusing every episode press (it gates `episodes::action`)
+        // for the rest of the session. See `stores/pump_wiring_tests` for the regression pin.
+        if crate::stores::metadata::pump_season() {
```
GLinnik21, line 2093: worth a line saying pump_detail() has to run first (it calls supersede_season() to invalidate a stale in-flight fetch) — the ordering here actually matters and nothing says so

rust-modules/src/stores/metadata.rs
```diff
@@ -218,3 +227,44 @@ impl<H: Host> Machine<H> for MetadataStore {
...
+    #[test]
+    fn the_route_unconditional_frame_update_pumps_the_season_landing_beside_detail() {
+        let src = std::fs::read_to_string(
+            std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("src/app/run.rs"),
+        )
+        .expect("read run.rs");
+        let detail_at = src
+            .find("crate::stores::metadata::pump_detail()")
+            .expect("run.rs must still pump the async detail landing every frame");
+        let window = &src[detail_at..(detail_at + 1500).min(src.len())];
```
GLinnik21: 1500 is a magic number and there's already ~1000 chars of comment sitting between the two calls — any future edit to that comment block turns this test red for no real reason

rust-modules/src/stores/metadata.rs
```diff
@@ -218,3 +227,44 @@ impl<H: Host> Machine<H> for MetadataStore {
...
+        assert!(
+            window.contains("crate::stores::metadata::pump_season()"),
```
GLinnik21: this just checks the string shows up somewhere in the window — it'd still pass if pump_season() got wrapped in a route gate, a dead cfg branch, or even just a comment

rust-modules/src/stores/metadata.rs
```diff
@@ -68,7 +68,16 @@
 //! Two route-unconditional per-frame landings sit beside `MetadataCmd` rather than inside it,
 //! because they are PUMPS (drain a mailbox, install what has landed) rather than requests a screen
 //! raises: [`pump_detail`], [`pump_season`], [`pump_alt_sources`] — called every frame regardless of
-//! route by `MetadataStore::step`'s `StoreEv::Pump` arm, exactly like every other store.
+//! route. `MetadataStore::step`'s `StoreEv::Pump` arm below folds all three through the same
+//! `note(StoreId::Metadata, …)` wrapper, but nothing ever SENDS a `StoreEv::Pump` to this store in
+//! production (only `Bridge`'s owned `BrowseStore` receives one, via `StoreWork::Browse`) — the
+//! arm is unreachable there. The actual per-frame callers are direct, unconditional calls in
```
GLinnik21, line 74: if this StoreEv::Pump arm is genuinely unreachable in prod now, why keep it around instead of deleting it (or making it the actual caller)?

rust-modules/src/stores/metadata.rs
```diff
@@ -218,3 +227,44 @@ impl<H: Host> Machine<H> for MetadataStore {
...
+
+#[cfg(test)]
+mod pump_wiring_tests {
+    //! `app/run.rs`'s route-unconditional per-frame block has no host fixture — `App` needs a real
+    //! SDL/GL context, so nothing in `cargo test` ever calls its `update()`. The one thing a host
+    //! test CAN pin is that the source still contains the call, the same idiom
+    //! `app::words::focusprobe_player_overlay_delegates_to_the_shared_overlay_word_function` uses
+    //! for the identical reason (a closure/sequence local to `run()` that no test can invoke).
+    //!
+    //! This is the regression test for the missing `pump_season()` call site: the legacy
```
GLinnik21: this is the third retelling of the same incident (module doc above + the run.rs comment + this) — could probably trim two of the three and let the commit message carry the history

rust-modules/src/stores/browse.rs
```diff
@@ -480,6 +480,13 @@ mod contract_tests {
             &first.browse.borrow().adapter,
             &second.browse.borrow().adapter,
         ));
+        // Drain any Hubs/Metadata/Search/Person notice a PRIOR test left dirty: those four are
+        // still process-wide globals (`stores::mod.rs`'s `NOTICES`), not owned per `Stores`
+        // instance, so `take_notices()` below would otherwise fold someone else's leftover
+        // bump into this test's precise assertion — the same idiom the other contract tests in
+        // this module already use before their own exact-equality check.
+        let _ = first.take_notices();
+        let _ = second.take_notices();
```
GLinnik21: since these 4 notices are process-wide globals, draining `first` already drains them for everyone — is `second`'s drain actually doing anything here?

GLinnik21 (approved):
Actually verified this on the real TV, not just reading the diff — deployed the pre-fix commit (8e96c159) to the debug install, opened The Office (6 seasons), switched Season 1 -> Season 2: tab highlighted but the episode grid stayed on Season 1 with the spinner stuck over episode 2 forever, and pressing OK on an episode did nothing. Redeployed this PR's branch, same repro steps, season switched instantly and pressing an episode started playback right away (S2E1 "The Dundies"). So yeah, this really was broken and the fix really fixes it \o/

Rest is a handful of nitpicks from a closer read, nothing blocking.

---
_Generated by [Claude Code](https://claude.ai/code)_

rust-modules/src/stores/metadata.rs
MrcRjs: Good question, and I thought about it but I'm going to leave it for a separate PR. 
Two reasons: it predates this PR (it's phase-4 scaffolding, not something I introduced), and PersonStore has the identical unreachable arm shape, so a real fix here is "decide what to do about this across both stores," not a one line edit to this file. The tempting option turning it into unreachable!() now that we've proven it's dead trades "silently redundant" for "hard panic on the TV if the reachability analysis is ever wrong somewhere I didn't check" which feels like the wrong risk to take inside a bugfix PR about a stuck spinner.

rust-modules/src/stores/browse.rs
MrcRjs: You're right, it does nothing. Traced it: take_notices() drains the 4 shared globals and that instance's own notice, but neither store had anything pending on its own half here (seeding never calls bump() the assertion two lines down already proves that, expecting gen 1 not 2+). So both calls only existed for the shared global drain, and the first one already covers that for everyone. Dropped second's call, fixed the comment.

rust-modules/src/stores/metadata.rs
GLinnik21: Agreed, defer it. Both stores in one follow-up is the right shape, and I'd rather not trade a silent no-op for a device panic inside a spinner fix. Open an issue so it doesn't get lost?

---
_Generated by [Claude Code](https://claude.ai/code)_

GLinnik21:
f29f8211 covers it. Good call verifying the new assert actually goes red on a re-indent and on a swap — that was the part worth proving. Resolving the rest; merging once host checks are green.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #115, opened 2026-09-17
Subtitle color: a gray ladder for the client-rendered caption

danzig666:
## Why

In HDR playback the panel maps graphics white far brighter than it does in SDR, and the pure-white caption is genuinely uncomfortable to read in a dark room. The only cure is less light, so this adds a brightness ladder rather than a colour wheel.

## What

The player's **Subtitles** menu gains a second section under the tracks, **Color**: `White`, `Silver`, `Light gray`, `Gray`, `Dark gray`, `Charcoal` — lightest first, `Row::checked` like the tracks above it. Picking one closes the panel (same as every other row), applies on the next frame, and persists.

- **`plex::session::SubtitleTone`** — persisted install-wide in the public preferences next to `playback_quality` (it is a fact about the panel, not the profile). Explicit on-disk spellings; an absent field or an unknown spelling is **white**, soft-parsed so it can never cost the credentials.
- **`theme::SUBTITLE_INKS`** — a new private `CAPTION_*` primitive family (exact 8-bit achromatic codes via `rgb8`) resolved by one public role, indexed by `SubtitleTone::index`. No raw literals at the call sites; the two `let white = [1.0, …]` carve-outs in `player_hud.rs` are gone.
- **Text cues** are inked with the tone; **image cues** (PGS/VobSub) are *tinted* by it — `fs_img` multiplies by the tint, so a bitmap keeps its authored colours and outline and only gives up light. The 0.85 black outline is untouched, which is what keeps the darkest rung legible.
- **`TrackCommit::SubtitleTone`** — the panel reports, the loop performs (`app::playback::commit_track` → `player::set_subtitle_tone`). A tone row deliberately does **not** produce `TrackCommit::Subtitle`: that one always republishes, and re-committing the track would re-burn a transcode. `tone_base` is captured when the table is built, so the track/tone split is by what was drawn rather than by re-asking `visible_subs`.
- Restored at the two places `route::restore_quality` is: boot, and the credentials handoff after a fresh sign-in.
- `docs/parity-gaps.md` — the three "pure white, no user control" entries get a *Partly landed* note. Size, position and background are still constants, and a server-burned subtitle is pixels this cannot reach.

No FFI, linkage, `dynlib!`, Starfish, ACB, curl or FFmpeg changes. No new telemetry event or schema field.

## Verification — and its limits

- `make check` green on this exact tree (lint + both unit-suite passes, 3,653 / 3,698), plus `cargo check --lib --no-default-features`, plus the armv7 cross-build, debug `.ipk`, `check-elf.sh` and `check-package.py` — run on my fork: https://github.com/danzig666/plx-native/actions/runs/35279163016 (that branch is this commit plus a fork-only workflow file).
- New host tests: the file contract (absent / damaged / every rung round-trips, survives `split_public` → `join_canonical` and `public_session`), the ink ladder (same length as the enum's ladder, strictly darkening, opaque, achromatic, darkest rung still above its outline), and `tone_at` (a track row is never a tone; past the ladder is `None`, not a neighbour — without the split a tone row would fall through to the `vis.get(..)` miss, i.e. switch subtitles **Off**).
- **Device:** installed and confirmed working on my own LG set — but as a hand **port onto `v0.6.6`** (branch [`subtitle-tone-v0.6.6`](https://github.com/danzig666/plx-native/tree/subtitle-tone-v0.6.6) on my fork), because that is what the released app is. That is one set and a short session, not a test matrix. **This `main`-based version has not been run on a television**: I have no rooted set or local NDK host (Windows x86_64), so the restructured overlay path (`PlayerOverlayScreen` → `PlayerReq::CommitTrack`) is host-tested and cross-built only. Image-subtitle (PGS/VobSub) tinting has not been specifically checked on device on either branch.

If you'd rather carry this on the `release/v0.6` line first, the v0.6.6 port above is ready to be turned into a PR against it.


GLinnik21:
Thanks! This landed as #158 (squash, you're credited as co-author). I couldn't push to your branch, so I opened a landing PR from this repo instead. It has your commit unchanged plus the review fixes, each listed in its description. Closing this one in favour of it.

PR #116, opened 2026-09-17
External (sidecar) text subtitles on direct play

danzig666:
## Why

A `.srt` next to the film shows up in the official Plex app but not here. The cause is the one `docs/parity-gaps.md` already records ("Sidecar (external) subtitle streams are unreachable on the direct-play path — i.e. on almost every item"): the demuxer only sees streams inside the container, so the track menu hides externals unless the item is already transcoding — and since they can't be picked, they can't even force the transcode that would burn them.

## What

- **`player::sidecar`** (new) — the missing third producer. Fetches the whole file from PMS once on a `task` worker, parses it (SubRip / WebVTT through one reader, ASS from its `Dialogue:` events, reusing `player::sub_text` for cleaning), keeps it sorted and answers by time. **Its own store, not `SHARED.sub_cues`**: that one is a 2 s / 512-cue window the demuxer refills, and a whole-file sidecar would not survive a backward seek in it. Generation-fenced so an answer for a pick the viewer already left is dropped. A failed fetch or an unparseable body says so **on screen for 6 s** in the caption's place, rather than silently drawing nothing.
- **`Client::sidecar_subtitle`** — asks `/library/streams/{id}?encoding=utf-8&format=srt` (both parameters are in `docs/plex-openapi.json`), which is what turns a Windows-1250 `.srt` or an `.ass` into something one parser reads. Falls back to encoding-only and then the bare key, because the conversion is the server's and the repo's own notes record `.vtt` answering 501. Only ever requests a `/library/streams/` path.
- **`ui::player_hud::draw_subtitles`** asks the sidecar first, then the embedded store, and takes `transcoding` from the frame's session: a transcode **burns** the selection, so the sidecar is silent there instead of doubling the line (e.g. a DTS audio pick mid-film).
- **`metadata::Stream`** carries the delivery `key`; `sidecar_renderable()` is a text-format **allow-list**, so image sidecars (`.idx`/`.sub`, `.sup`) stay transcode-only rows rather than being offered and then blank.
- **Track menu** — renderable sidecars are listed on direct play with an `EXTERNAL` badge. `TrackCommit::Subtitle` gains `sidecar_key`; the panel still only reports, and `app::playback::commit_track` selects/deselects the sidecar beside the unchanged `commit_subtitle_selection`.
- **`route::apply_plan`** — a server-selected sidecar is restored on a direct-play start (the external twin of `pick_dp_subtitle`, which is unchanged and still returns `None` for it). It rides outside the `Plan` on purpose: no demuxer ordinal, no burn, so nothing in the route contract moves; the menu's checkmark falls back to `sidecar::selected_stream_id`.
- Docs: the parity-gaps entry and `player/CLAUDE.md`'s subtitle bullet.

No FFI, linkage, `dynlib!`, Starfish, ACB, curl or FFmpeg changes. No new telemetry.

## Verification — and its limits

- `make check`, `cargo check --lib --no-default-features`, the armv7 cross-build, debug `.ipk`, `check-elf.sh` and `check-package.py` are green for this exact commit on my fork: https://github.com/danzig666/plx-native/actions/runs/35282006996 (that branch is this commit plus a fork-only workflow file).
- New host tests (all pure): SubRip with BOM/CRLF/tags/counter lines, WebVTT header + cue settings + optional hours, ASS tenth-field text with commas and centiseconds, sorting and dropping of unusable cues, non-subtitle bodies (an HTML error page, binary) yielding no cues, timestamp edge cases, and the time lookup across forward and backward "seeks" including overlap and the exclusive end.
- **Device:** confirmed working on my LG set with **Hungarian `.srt` sidecars** (they display correctly, which suggests the server-side UTF-8 conversion does what the spec says; I did not check which of the three request forms answered) — but as the same change **ported onto `v0.6.6`** ([`subtitle-tone-v0.6.6`](https://github.com/danzig666/plx-native/tree/subtitle-tone-v0.6.6) on my fork), since that is the released app. **This `main`-based version has not been run on a television**; the differences from the tested port are the explicit `PlaybackSession` plumbing (`transcoding` passed into the draw, `TrackCommit` carrying the key). One set, one server, `.srt` only: `.ass`/`.vtt` sidecars, the fallback request forms, the restore-on-next-play path and the mid-film direct→transcode hand-off are host-tested or reasoned, not device-verified.
- Known gap left alone: there is still no subtitle delay control for an out-of-sync sidecar (its own parity-gaps entry).

Independent of #115 (subtitle color); the two touch `track_menu.rs` in different places and will need a trivial merge in whichever lands second.


GLinnik21:
Thanks! This landed as #167 (squash, you're credited as co-author). I couldn't push to your branch, so I opened a landing PR from this repo instead. It has your commit unchanged plus the review fixes, each listed in its description. Closing this one in favour of it.

Issue #107, opened 2026-09-17
Some cover art never load


airecken:
Some cover art never load properly and just a blank black box persist.  Cover art is seen fine on other Plex clients.

LG Tv is 55C9AUA running firmware 5.50.  Native Plex app is uninstalled.

GLinnik21:
Thanks, I’ll try to look into this. Could you give me a little more data to work with?

If possible, please share:

* a screenshot showing the missing artwork
* a few examples of affected movies/shows
* logs captured after reproducing the issue

That should give me something to trace, since right now I can’t tell whether the image request is failing, the response is invalid, or something goes wrong while loading/rendering it.

airecken:
Captured the logs from the affected LG 55C9AUA (firmware 5.50), running PlxNative 0.6.6 on webOS TV 4.10.2. I have a full developer bundle with the raw app logs, metadata, and a reduced failure timeline.

The useful sequence from the current run is:

```text
plex: server slot 0 registered at https://<host>
hubs: source 0 ok — 2 shelves, 11 in CW
hubs: landed — 35 items, 3 shelves
plex: server slot 0 re-pointed to <addr>
security: refused plaintext PMS credentials; HTTPS required
pms: server 0 info unavailable (GET / failed) — subscription stays unknown
hubs: source 0 FAILED (attempt 1) — retrying in 2s
posters: art fetch FAILED on server 0 - no usable response - refused/timed-out connect, or a status outside 2xx (tiles stay skeletons; further ones like this are silent)
plex: server slot 0 re-pointed to https://<host>
auth: source 0 endpoint refreshed after transport failure
hubs: source 0 ok — 2 shelves, 11 in CW
hubs: landed — 35 items, 3 shelves
```

Later in the same run there is also `curl rc=6 — could not resolve host` and `curl rc=28 — timed out`. There are no `img: decode-none` lines, stderr is empty, and the crash sidecar contains only the binary build ID.

My read of v0.6.6 is that this is an endpoint-recovery/poster-cache interaction. The poster worker changes the slot to `P_FAILED` on any unusable response. `lookup` then continues matching the same `(ServerId, built path)` and returns texture 0, so the tile stays a skeleton until LRU eviction. Re-pointing the same server slot back to HTTPS creates a fresh client and bumps the roster generation, but does not invalidate/re-arm matching failed poster slots. The source comment in `posters.rs` describes that persistence explicitly.

[...]

PR #134, opened 2026-09-18
Posters: retry a transient fetch failure instead of burning the slot (#107)

danzig666:
Fixes the "never retried" half of #107.

## Why

A poster fetch that came back with nothing marked its slot `P_FAILED`, and `lookup` kept matching that key and answering "no art" for as long as the slot lived (the module's own doc says so: *a skeleton until the LRU walks back to the slot*). Nothing distinguished "the server said 404" from "the network was briefly unusable", so a cover that missed **once** stayed blank for the session.

On v0.6.6 there was a miss every release build made: the address race activated the plaintext `http://` twin first, `http::credential_transport_allowed` refused to put the token on it, and every art request in that ~100 ms window failed as `Transport`. The debug build allows plaintext tokens, which is exactly why the reporter (and I) saw it on the stable id and not the debug one. #104 closed that window on `main` — but a refused or timed-out connect, a DNS miss (`curl rc=6` / `rc=28` in the issue's logs) or a 5xx still burn a slot the same way today.

## What

- **`Client::fetch_built_outcome`** keeps *why* there are no bytes — `ArtFetch::{Bytes, Status(i32), NoResponse}`; `fetch_built` is now a fold over it.
- **`P_RETRY`** (poster adapter): a transient failure — no response, 5xx, 401/403/429, an empty body, or no client registered for the id yet — parks the slot instead of burning it. The first **draw** that finds it schedules a wait on `app::clock` (1 → 2 → 4 → 8 → 16 → 30 s), and the first draw after the wait re-queues it as `P_WANT`. The worker only records the outcome and never reads the loop's clock (the `wall` gate stays clean).
- **Still final:** 400/404/410/415 and bytes the decoder rejects stay `P_FAILED`, so the negative cache for artwork that genuinely does not exist is unchanged.
- `P_RETRY` is settled for the LRU like `P_FAILED`, so a tile that scrolled away holds nothing, and nothing retries off screen (the re-queue is on `Touch::Draw` only).
- The once-per-(cause, server) log line now says a 4xx is final and everything else is retried, instead of "tiles stay skeletons".

## Verification

- `make check`, `cargo check --lib --no-default-features`, the armv7 cross-build, debug `.ipk`, `check-elf.sh` and `check-package.py` are green for this exact commit on my fork: https://github.com/danzig666/plx-native/actions/runs/35368574712 (that branch is this commit plus a fork-only workflow file).
- New host tests: `only_a_final_status_is_final` (the transient/final split, including the boot window's `NoResponse` shape), `the_retry_backoff_doubles_from_one_second_and_caps`, `a_retry_slot_is_due_when_its_wait_is_over` (unscheduled is not due, the tick comparison wraps, `P_FAILED` is never due), and the LRU test now also covers `P_RETRY` as settled.
- **Device:** the same change ported onto the `v0.6.6` line (branch [`subtitle-tone-v0.6.6`](https://github.com/danzig666/plx-native/tree/subtitle-tone-v0.6.6) on my fork, using `Instant` there since that line has no clock gate) is running on my LG C3 (webOS 23) on the stable id, where the missing covers reproduced reliably before and appear now. This `main` version — the `app::clock` scheduling in particular — has been host-tested and cross-built, not run on a television.


GLinnik21:
Thanks! This landed as #159 (squash, you're credited as co-author). I couldn't push to your branch, so I opened a landing PR from this repo instead. It has your commit unchanged plus the review fixes, each listed in its description. Closing this one in favour of it.

PR #135, opened 2026-09-18
Player: a seek no longer throws the playbar to the left edge, nor flashes the old position

danzig666:
Two independent defects behind one report — "the seek bar jumps when I seek" — both on an LG C3 (webOS 23), where in-place seeking is disabled (`INPLACE_SEEK_OK` is false under `VP_EXPORTED`) and so **every seek is a reload**. They should be visible on any webOS 5+ set; the dev set seeks in place and never takes the first path.

## 1. The playbar sat at the far left until the file was reopened

The tell was that the **clock stayed correct** while the bar went to zero. The clock is the position alone; the bar is position ÷ duration.

`reload_at` → `teardown(for_reload = true)` → `SHARED.reset_session()` zeroes `duration_ns`. `arm_seek` then seeds `playpos_ns` with the target (so the clock is right at once), but the duration's only writer on direct play is the demuxer's open in `ff.rs`, a few hundred ms later. Until then `frac = dispos / dur` with `dur == 0` is `0.0`.

The duration is a fact about the **file**, and a reload plays the same file, so `Shared::reset_session_for_reload` keeps it across the reset. A real stop still zeroes it — the next item is a new file.

## 2. One frame of the pre-seek position on every seek

Sequence seen on the panel: scrub preview → **old position for a flash** → target.

`request_seek` sets `SHARED.seeking` at the press, but `pb_state` is only republished at the end of a pump pass. A reader in between still saw `Playing`; the HUD freezes the playhead on `seek_display_ns` only while `loading()`, and the scrub preview had already been cleared, so it fell through to `livepos` for a frame.

`player::state()` now derives `Seeking` from the flag, in the same precedence the pump's own ladder gives it (a seek outranks frames). It already derives `Resolving` and `Error` in the reader for the same "single writer, but the reader must not lag it" reason.

## Verification

- `make check`, `cargo check --lib --no-default-features`, the armv7 cross-build, debug `.ipk`, `check-elf.sh`, `check-package.py`: green for this exact commit on my fork — https://github.com/danzig666/plx-native/actions/runs/35375188963 (that branch is this commit plus a fork-only workflow file).
- Two host tests in `player::seek_hud_regressions`:
  - `a_reload_keeps_the_files_duration_and_a_stop_does_not` — a differential: the old reload path called `reset_session`, for which the first assertion reads 0. Also asserts the session's own state (`playpos_ns`) is still cleared.
  - `a_requested_seek_reads_as_seeking_before_the_pump_republishes` — arms the flag through the real `request_seek`, deliberately: my first version set the flag by hand and `tests/test_harness.py`'s `test_the_flag_is_armed_in_exactly_one_place` correctly refused it.
- **Device:** both fixes are running on my LG C3 as ports onto `v0.6.6` ([`subtitle-tone-v0.6.6`](https://github.com/danzig666/plx-native/tree/subtitle-tone-v0.6.6) on my fork), where each symptom reproduced on every seek — D-pad scrub and Magic Remote click alike — and is gone. They were fixed one at a time, which is how the two were told apart: after (1) the bar stopped going to zero and the one-frame flash of the old position was what remained. This `main`-based commit itself has not run on a television; the two edits are the same lines in the same functions, the one difference being `state()`'s `ps` parameter.

Not included: on my fork BACK with the transport up hides it rather than leaving playback. That is a behaviour preference rather than a bug, so I left it out.


GLinnik21:
Thanks! This landed as #157 (squash, you're credited as co-author). I couldn't push to your branch, so I opened a landing PR from this repo instead. It has your commit unchanged plus the review fixes, each listed in its description. Closing this one in favour of it.

PR #149, opened 2026-09-19
Honour a show's audio and subtitle language settings

danzig666:
## Why

Plex lets you set, per show (**Edit → Advanced**), a *Preferred audio language*, a *Preferred subtitle language* and a *Subtitle mode*. The official clients honour them; this one never read them. Reported on my set: a series set to Hungarian opened in English every time.

The cause is in `pick_dp_audio`'s ladder. The setting lives on the **show**, and an episode's part carries nothing of it. When the preferred dub is also the file's `default`-flagged track, which is common for dubbed releases, rung 1 (a server selection that differs from the default) cannot tell it apart from "nobody chose". So rung 2's built-in English preference wins.

## What

- **Reading the settings.** `Client::show_language_prefs` reads `audioLanguage`, `subtitleLanguage` and `subtitleMode`:
  - it asks `/library/metadata/{show}?includePreferences=1` first. That is what Plex's own clients use for the Advanced dialog, but the parameter is **not** in the vendored `docs/plex-openapi.json`;
  - so it falls back to the one read the spec *does* document carrying these settings, `/library/metadata/{id}/tree`, whose container holds a `Setting[]` (the spec's `show` example lists all three, with their `enumValues`);
  - the fallback is only asked when the first answer carries none of the three.

  `ShowLangPrefs::from_settings` parses either response. Empty values and `-1` are "Account default", which lives on plex.tv; this client does not read it and treats it as unset.
- **Knowing the show.** `metadata::PlayingItem` now carries `show_rk` (`grandparentRatingKey`), from both the cached and the fetched builder.
- **Using them.** `route::plan::build_stream` asks once per *episode* play, then applies three rules.
  - **Audio: `pick_dp_audio_pref`.** A direct-playable track in the show's language is chosen right after rung 1 and before the English preference and the default flag.
    - Rung 1 still wins: a *different* track picked for this episode on another client is more specific than a show-wide default.
    - A preferred track that cannot direct-play (DTS/TrueHD) does not force a transcode. This is the same reasoning rung 1 already documents, and it is logged.
  - **Subtitles: `pick_dp_subtitle_pref`.** Any existing server selection still wins, and any `selected` subtitle, a sidecar included, keeps the show default out. Otherwise:
    - mode **2** (always enabled) turns on an embedded track in the show's subtitle language;
    - mode **1** (shown with foreign audio) turns one on only when the audio that will play is in another language;
    - mode **0** (manually selected) and **-1** (account default) change nothing.

    Among matching tracks it takes a full non-SDH track first and a forced one last. It is direct play only, as before, since turning one on for a transcode means a burn. The decision is made **once** and used for the MDE handshake (`mde_subtitle_id_of`), the Original candidate and the direct-play plan, so the three cannot disagree.
  - **Language matching: `lang_matches`.** It maps the BCP-47 preference (`hu-HU`) to ISO-639-2 stream codes, accepting both the B and T spellings (`ger`/`deu`, `fre`/`fra`), for every language the setting's `enumValues` offer.

One extra small GET per episode play; nothing on movies.

## Verification

- `make check`, `cargo check --lib --no-default-features`, the armv7 cross-build, debug `.ipk`, `check-elf.sh` and `check-package.py` are green for this exact commit on my fork: https://github.com/danzig666/plx-native/actions/runs/35435440406 (that branch is this commit plus a fork-only workflow file).
- New host tests in `plan_track_selection_tests.rs`:
  - `a_shows_preferred_audio_language_is_honoured` is differential: plain `pick_dp_audio` still picks English on the same tracks. It also covers the default-English case, a non-DP preferred track, and a per-episode pick outranking the show;
  - `a_shows_subtitle_settings_choose_the_starting_subtitle` covers all four modes, full-over-forced, sidecars never picked, and an existing selection winning;
  - `preferred_language_tags_match_both_three_letter_spellings`;
  - `show_settings_are_read_out_of_a_setting_list`.
- **Device:** confirmed on my LG C3 (webOS 23) against my own server. A series set to Hungarian now opens in Hungarian, and its subtitle settings apply. That was tested as the same change ported onto `v0.6.6` ([`subtitle-tone-v0.6.6`](https://github.com/danzig666/plx-native/tree/subtitle-tone-v0.6.6) on my fork), where the port additionally caches the last show's settings and can pick an external `.srt`, since that branch also carries #116.

  **This `main`-based commit has not run on a television.** It also has not been checked which of the two endpoints my server answered: the fork logs the outcome, not the endpoint.

If #116 (external subtitles) lands, `pick_dp_subtitle_pref` is the one place to add a sidecar tier after the embedded ones.


GLinnik21:
Thanks! This landed as #160 (squash, you're credited as co-author). I couldn't push to your branch, so I opened a landing PR from this repo instead. It has your commit unchanged plus the review fixes, each listed in its description. Closing this one in favour of it.

Issue #165, opened 2026-09-19
No reliable way to select between multiple libraries of the same type


GLinnik21:
## Goal

Viewers with more than one library of the same type (two TV Shows libraries, HDR and FHD movie libraries, kids and adult libraries) can reach each of them from the Movies and TV Shows tabs, and disabling a library in Settings is honoured.

### GitHub
- #68 (sub-issue) — two TV Shows libraries; the TV Shows tab shows only the anime one, and disabling it in Settings doesn't change which one opens ([comment](https://github.com/GLinnik21/plx-native/issues/68#issuecomment-5559979565)). TV, webOS and PlxNative versions not stated.
- #100 (sub-issue) — Films / Films HDR / TV / TV HDR are merged into two lists; the Library picker is missing, only Sort and Filter show. A [second user](https://github.com/GLinnik21/plx-native/issues/100#issuecomment-5687909292) reports the same with two movie and two TV libraries. TV, webOS and PlxNative versions not stated; the maintainer calls it a known 0.6.x gap.

### Discussions
- [#71](https://github.com/GLinnik21/plx-native/discussions/71) — anime and anime-movies libraries are missing from the tabs (findable in Search). [After updating to 0.6.6](https://github.com/GLinnik21/plx-native/discussions/71#discussioncomment-18466977): "no change … still only see 2 libraries". TV and webOS not stated.
- [#97](https://github.com/GLinnik21/plx-native/discussions/97) — the original HDR/FHD split-library thread; closed and moved to #100.

### Reddit
- [Two movie and two series libraries (kids vs grown-up); each tab opens only one, the others are reachable only via Search; disabling the unwanted ones still opens the wrong one](https://www.reddit.com/r/webos/comments/1wj44l2/i_built_a_native_plex_client_for_lg_webos/paglm1p/) — TV, webOS and PlxNative version not stated.

### Plex Forum
- None found in the three threads reviewed.

## Known failure modes

1. **No way to switch between libraries of the same type** in the Movies / TV Shows tab (#68, #100, #71, Reddit).
2. **The tab opens one library the viewer didn't choose** (#68, Reddit: "it would still choose the wrong one").
3. **Disabling a library in Settings doesn't change what the tab opens** (#68, Reddit). #91 documents that in 0.6.x the setting only governs Home.
4. **Libraries reachable only through Search** (#71, Reddit).

## Open questions

[...]

Issue #100, opened 2026-09-15
Library picker is missing in Movies and TV Shows


GLinnik21:
### Description

When multiple Plex libraries of the same type are enabled, the **Library** picker is missing from the Movies and TV Shows screens. Only **Sort** and **Filter** are shown, so content from the enabled libraries appears merged with no way to browse a single library.

Reported setup:
- Films
- Films HDR
- TV
- TV HDR

All four libraries are recognized and selected during initial setup, but the per-library picker is not available afterward.

### Expected behavior

Movies and TV Shows should show the existing **Library** picker, allowing the user to switch between enabled libraries of that type.

### Discussed in https://github.com/GLinnik21/plx-native/discussions/97

<div type='discussions-op-text'>

<sup>Originally posted by **Dal78** September 14, 2026</sup>
Within my plex server i maintain an HDR/4K library and a separate FHD library for both movies and shows.

Currently plx-native merges those 4 libraries into two large lists however it is difficult to select which format to play.

A method would be desirable to either seperate or filter them</div>

kjettern:
+1 here! I also maintain 2 separate libraries for movies and tv-shows.. I would like to see them split apart without filtering.. btw love your plx-native for its speed!

Dal78:
If you want any debug info on this just let me know what you want and the procedure to get if its not common sense.

GLinnik21:
Good catch — that's a known gap in 0.6.x (the picker only handles multiple servers, not multiple libraries per section); I actually pulled that old menu in 0.7.0 and I'm hoping to bring back a proper per-library picker in its place!

GLinnik21:
Fixed on `main` by #180. The Movies and TV Shows screens now show per-library pills when multiple libraries of the same type are available, and remember the choice per profile. This will be part of 0.7.0 and is available in nightly builds. Closing as completed.

Dal78:
Tested the nightly and it works well.

My one comment would be while the "Recently Added on Library" lists correctly within each pill selected area on the root home screen recently added only shows for the first library of that type and not the second.

So on native plex app i will have "Recently added on TV" and "Recently added on TV HDR"

Issue #68, opened 2026-09-06
Not all my TV Shows libraries are being displayed


astrologg:
Hello, 

I have two TV Shows libraries, one for TV Shows and Another for japanese Animes. In the settings menu, I can confirm the app sees my two differents TV Shows libraries. But when I click on "Tv Shows" on top, only my "Animes" are being displayed.

GLinnik21:
Thanks for the report! The TV show libraries are actually kept separate — you can switch between them using Libraries in the filter pills.

I agree that merging them into a single view could be more intuitive. For now, though, I’m going to keep them separate to match Plex’s per-library structure, especially since the next update will add Plex recommendations, which are also provided per library.

I’ll close this as not planned for now, but I’ll look into making the library switcher more obvious.

GLinnik21:
I actually considered putting individual libraries directly in the top tab bar at first. The problem is that some users can have quite a few of them, so I ended up grouping the top-level tabs by Plex library type (Movies, TV Shows, etc.) and keeping individual libraries nested inside.

There might still be a better way to make that hierarchy more obvious, though.

astrologg:
Thanks for your quick answer. The only problem is that I cannot switch between the libraries from there. Even if I disable my "Anime" library in the settings, only my "Animes" library is populated under "TV Shows." And I don't have any option to switch the library from there.

GLinnik21:
Ah, got it — I misunderstood the original report then. If you have multiple TV libraries, you should definitely be able to switch between them there.

I’ll reopen this and take a look. Thanks for clarifying!

GLinnik21:
Fixed on `main` by #180. Movies and TV Shows now show a per-library selector whenever more than one eligible library of that type is available, so both TV libraries can be reached directly. This will be part of 0.7.0 and is available in nightly builds. Closing as completed.

PR #180, opened 2026-09-20
Library: a selector that counts, and a managed profile's own family server graded as the household's

GLinnik21:
Five reported UI regressions, and — following the "merged libraries" reports down to their
mechanism — the Plex Home hole that made two of them come back for a managed or Guest profile.

## The five

1. **A Guest saw the legacy `Library · <name> ⌄` chip instead of the pill selector.** The selector
   now appears when, and only when, more than one eligible library of the type exists, and each
   pill is the same `TabPill` the Season row uses, suffixed with an owner's handle where there is
   one. The singleton chip and its Sources escape hatch are gone.
2. **A grid tile in "All" never played its unfocus animation.** The rect and the visual treatment
   were computed in two places; `library::parts::tile_scale` is now the single source of both.
3. **"Stats for Nerds" sat above the quality rungs.** `more_menu::rows_for` emits quality first.
4. **Two Settings details were too long to read at a glance.** Shortened.
5. **Pushing a nested Settings screen landed focus in the wrong place.** `FocusTarget` gained
   `FirstInGroup(GroupId)` and `NavStack::fresh()` names the group it means, so a freshly pushed
   screen seats on the first element of its own band. BACK still restores the remembered cursor.

## Why the libraries looked "merged", and what was still open

The reports (#68, #100, #165) are against 0.6.x, and nothing merged there — navigation did. That
build keyed the tab strip by TYPE (`TAB_KINDS`, two entries), picked one section per type on raw
ownership with no favourite filter and no remembered choice, and its own banner said the Settings
toggle governed Home alone. 0.7 closed all three with the favourite filter, the remembered
per-profile choice and the library selector — **for an account owner.**

plex.tv grades a Plex Home managed profile's own family server `owned:false`, with the admin in
`sourceTitle`, exactly as it grades a friend's share. So for a managed or Guest profile nothing was
owned, and every ownership-derived rule degenerated back to the 0.6 state: the pin defaults could
not prefer the house's libraries, the tab's tiebreak had nothing to break the tie with, and Search
credited the household's own server to the admin by name.

`home` and `ownerId` are now carried as evidence from `Resource` through `ServerFacts`,
`SourceRef`, `CandidateActivation`, `BrowseSource` and `ScopeSource`, and `is_household` grades a
grant against the Plex Home roster rather than against `owned` alone. `pins::default_on`,
`browse::section_of_kind` and the Search scope line read that verdict. An account owner is
unaffected by construction — `owned` is the predicate's first clause.

Household grading does not, and should not, tell Anime from TV Shows on one family server. The
library selector does that, and it is not ownership-gated.

## The roster lands late, so the table has to be re-resolved

`/api/v2/home/users` arrives on its own schedule and changes no server fact, so
`discovery_needs_pump` compares the live verdict against the stored one, and `sync_roster_owned`
re-resolves the whole pin table when a source changes sides: a source moving reclassifies
`household_type` for every library of its kind on every other source at once.

Re-resolving has to know which rows are a real answer, so `pins::record` takes one `Option<bool>`
per library and `BrowseCmd::ApplyPins` carries the rows the viewer **answered** rather than the
whole visible draft. Two failures this closes, both found in review:

- Inferring "answered" from "differs from the live pin" loses the row a roster correction moved
  onto the value the viewer had just chosen — it records nothing, and a later reclassification
  flips it back against an explicit choice.
- Recording answers without re-resolving left the screen showing a floor-raised `On` while the
  record kept `Off`. Visible to a plain account owner, and the row vanished at the next restart
  with no user action. The commit now reconciles through the same path the reclassification uses.

## Verification

- `test result: ok. 3993 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out`
- `cargo +nightly check --lib --no-default-features` clean; clippy clean.
- Each fix has a regression test that was observed failing first. The selector carries a
  20-cell tier x state matrix and a 5-roster ownership table asserting that a managed/Guest
  profile draws the same pills, with the same bare handles, as an owner.

Not covered here: pixel output and LG text rasterization, which need the television; and the two
empty-credit readers (`pms::roster`'s Home grouping, `metadata::alt_copies`' "This account" row),
which are pinned by a test that asserts the current wrong output so it speaks up the day it moves.


GLinnik21:
@codex review

PR #181, opened 2026-09-20
feat(i18n): runtime UI translation with tv-locale detection (Spanish included)

gnacho:
This adds a runtime translation engine for the whole visible UI. English stays the compiled-in source: every draw site keeps its English literal, which is also the lookup key, and a static Spanish table maps it to the translation. The language is read once at boot from the television's own UI locale (/var/luna/preferences/localeInfo), before the first frame, with no in-app override; any failure to read the locale falls back to English, and an untranslated key renders as its own English string rather than going empty. Fixed labels drawn through C-string APIs pick between two static spellings, so the draw path stays allocation-free. The Spanish table covers the full visible UI today, including the long legal and consent documents, which live beside the English ones as Spanish bodies picked by the locale. A host test pins every table entry against an empty or identity translation, and another flips the language atom and reads the player captions in both languages.

Honest caveats: translations are keyed by the English source string, so an edit to an English literal silently untranslates it until the table catches up (a pin-list test covers the table's own entries, not every draw site). Device behavior was not verified on a television; verification is the host suite below.

Verified on the host toolchain (nightly-2026-07-02): the full make check host suite passes (3999 + 4046 Rust tests, 85 storage-bin tests, all Python and C gates) except three build-gc tests that also fail on unmodified main on this machine because it has no lsof; cargo check --lib --no-default-features passes; cargo test --lib i18n passes 4/4. The ARM/NDK build was not run here and is left to CI.


GLinnik21:
Thanks for working on this. I can see that a lot of effort went into translating the whole UI.

I’m going to close this for now, though. The approach I have in mind is separate JSON files with scoped, stable keys, rather than using the English text as the key and keeping translations in Rust. Moving this PR to that approach would basically mean rewriting it.

I haven’t properly designed that system yet, and localization isn’t something I can focus on right now, so I don’t want to ask you to spend even more time on it without a clear spec. The Spanish translations here should still be useful when I get to it. Thanks again!

PR #182, opened 2026-09-20
posters: one notification for art the render cache cannot keep resident

GLinnik21:
Closes part of #107.

## The defect

The texture cache and the poster source held two answers to the same question and never reconciled them.

When residency crossed `TEX_RESIDENT_BYTES_MAX` (44 MiB) the cache dropped the least-recently-used texture and told nobody. The source's slot stayed `P_READY` and kept answering "I have this image", while `resolve` returned absent and the tile drew its skeleton. Nothing put the slot back to `P_WANT`, so the placeholder stayed **until the source LRU happened to recycle that slot for a different key** — which, for a tile you keep looking at, is never.

A systematicity audit found the **same defect wearing a second hat**: the cache's own `failed` set, reached through rejection instead of pressure. A decode error put the key there, and it cleared only on a successful `accept` that no source re-arm could ever produce.

## The change

The `failed` set is gone. Eviction and rejection now say one sentence — `Source::unresident(key)` — drained after the cache borrow is released, so the callback may touch application state without coupling the library half to it. The source parks both cases in a new `P_EVICTED`.

Review then found that re-arming an evicted key was unbounded: under sustained byte pressure key A is evicted, its probe re-arms it, its upload evicts key B, B's probe re-arms B, and so on. Two mitigations answer it, and they cover different halves:

- **Only a `Touch::Draw` probe re-arms.** A prefetch probe of an evicted slot returns `Warm::Known` and leaves it dormant — speculation cannot protect what it restores. The cost is real and is not hidden: an off-screen backdrop that lost residency now returns only when it is drawn.
- **A per-slot cooldown.** A key's first eviction re-arms on its next draw immediately. A key evicted *again* inside a 10 s window escalates a bounded backoff (250 ms → 8 s cap) before its next draw can re-arm, and one expiry requests exactly one redraw.

Neither makes an oversized *visible* working set converge, and the prose does not claim otherwise. They bound the rate, not the duration.

Recovery is also not cheap, which is why it is rate-limited: `imgcache`'s disk tier holds plex.tv avatars only, so an evicted poster pays a full network refetch, decode and upload.

The `ui/` ↔ `app/adapters/` split is unchanged — the cache still knows only an opaque key and never what it denotes.

## Verification

Gates on the committed tree, working tree empty before and after:

```
test result: ok. 4012 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 4059 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok.   85 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

`--no-default-features` (the shipping set, not covered by `make check`): clean. `ci/check-deps.sh`: all gates green.

**On the television** (rooted webOS 4.5, guest boot, muted), with the two counters this PR adds:

```
posters: residency lost=18 rearmed=4
```

Both halves of the mechanism fire on hardware: eighteen evictions under ordinary navigation, four re-arms. Artwork was complete throughout — no skeleton on any grid tile or detail page at that state.

One measurement worth recording because it invalidates an obvious check: **scrolling the poster grid produces no evictions at all.** `CACHE_CAP` equals the source's `PT_CAP` (both 64), so the cache can never hold more keys than the source has slots, and 64 posters is ~24 MB against a 44 MiB ceiling. Byte pressure needs backdrops and hero logos. "I scrolled the library and saw no skeletons" measures nothing here.

An earlier revision of this description said six 1920×1080 backdrops break the ceiling; backdrops are requested at 1280×720 (≈3.7 MB), and the pressure comes from backdrops plus hero logos plus the shelves behind them.


GLinnik21:
@codex review

Worth being adversarial about two things specifically:

1. `Source::unresident` is called from both `accept` and `prepare`, in each case after the cache borrow is released. Is there any path where a key is queued as unresident and then never drained — a frame that does not present, a shutdown, a `free`/`drain_all` racing the queue?
2. `P_EVICTED` is a settled state and so a valid LRU victim. Can a key oscillate — evicted, re-armed by a draw, evicted again — under steady byte pressure from large backdrops, and if so is that bounded by anything?

GLinnik21:
@codex review

PR #183, opened 2026-09-20
plex: a stored server becomes current only if its origin may carry the credential

GLinnik21:
Closes part of #107.

## The defect

Discovery already refused to activate a plaintext candidate. Restoring a **stored** session did not.

Installation gated on `SourceRef::dialable`, which is well-formedness — the port fits an `i32`, the address parses — and says nothing about whether a token may ride that scheme. A legacy entry holding only `address` + `port` has its origin synthesised as `http`, and `servers::activate` published it as the current credentialed client without a second look.

The transport then did its job and refused every token-bearing request. So the session was not merely artwork-less: **hubs, metadata and library browsing all failed together**, and the only trace was one `refused plaintext PMS credentials; HTTPS required` line. That is exactly the sequence in the log attached to #107.

This is reachable in a shipping build. `CredentialPolicy::build()` is `HttpsOnly` there — but that only made the transport *refuse*; it never stopped such a client from becoming active.

## The change

`CredentialPolicy` is unchanged and remains the single `cfg!` for this rule. What changes is **where it is asked**.

A `CREDENTIAL_ELIGIBLE` bitmask now sits beside `ACTIVE` and is written in `activate()` — the one point every registration path reaches: stored primary, roster, controlled bootstrap, discovery activation and endpoint repair alike. `set_current` refuses a slot that is not in it.

An ineligible origin is **not dropped**. A user whose only stored entry is plaintext would otherwise lose their server with no way back, so the entry stays active as recovery metadata with a blank token, visible to Sources and to the rediscovery loop; a later eligible HTTPS registration re-points the same slot, clears the verdict and restores ordinary activation. Developer builds under `AllowPlaintext` behave exactly as before.

`client()`'s doc no longer claims an unchanged contract: its precondition genuinely moved from "something installed" to "something credential-eligible installed". No reachable panic was found — production paths use `client_for`/`client_opt`, and direct `client()` calls are test support — but the stale claim was worth correcting before someone relies on it.

## Verification

Gates on the committed tree, working tree empty before and after:

```
test result: ok. 3999 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 27.49s
test result: ok. 4046 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 27.98s
test result: ok. 85 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.15s
```

`--no-default-features` (the shipping set, and the feature set this defect lives in): clean. `ci/check-deps.sh`: all gates green.

New tests cover the plaintext stored source under `HttpsOnly`, the legacy `address`+`port` synthesised-HTTP path, the unchanged `AllowPlaintext` behaviour, a cold boot carrying one HTTP-only stored source through install plus boot activation, and recovery of retained insecure metadata to HTTPS.

## Relationship to #182

Independent. #107 has two separate causes and this is the other one; neither PR depends on the other, and the files do not overlap.


GLinnik21:
@codex review

Two things worth attacking directly:

1. `CREDENTIAL_ELIGIBLE` and `ACTIVE` are separate atomics written in `activate()`. Is every read of the pair consistent, or is there an interleaving where a slot is ACTIVE-and-not-yet-eligible (or the reverse) that a concurrent `current()`/`client_for` could observe?
2. The recovery story is the part with no device coverage: an ineligible origin stays active with a blank token. Is there any path that can still attach a token to it, or that treats the blank token as a valid grant rather than as 'not usable yet'?

PR #186, opened 2026-09-21
Session save failures now carry an errno and a photographable read-out

GLinnik21:
A save failure told a reporter with no shell exactly nothing: `write_atomic` returned a bare `bool`
and discarded every errno, so a total fallback failure logged only
`could not persist to ANY candidate path` — indistinguishable from a server-side authorization
problem, and useless for deciding whether the jail refused the write or the path was never there.

`write_atomic` and `create_private_temp` now return a typed `WriteFailure` carrying the OS errno
where there is one, and `save_legacy_fallback_locked` logs one line per failed candidate: the path,
the parent directory's uid/gid/mode, and the errno or refusal class. It iterates whatever
`paths::session_candidates()` returns rather than assuming a fixed count, so it keeps telling the
truth as that list changes. This is the difference between EACCES and EROFS being visible at all —
on the reporting set they mean two entirely different things, one a permission bit and one a
read-only mount.

The failure read-out's existing Details card also names the persistence class, key-manager stage
and service error code. Those three were already defined on `telemetry::incident::IncidentContext`
and already sent to Sentry; they were simply never shown to the person looking at the screen. Both
projections now read the same typed evidence through one `storage_evidence_line` helper, and an
absent field renders as a fixed `unknown` rather than vanishing, so the line's shape can never
signal more than it knows. No new visual primitive, no devtrigger, no debug-flavour gate — it sits
behind the affordance that already ships.

`PRIVACY.md` and the in-app privacy policy named every sign-in-report field `event_body` could emit
except those three, which have been shipping silently. Their prose is corrected to match what the
serializer has always sent. That is a notice correction, not a widening: no consent scope changes
and `POLICY_VERSION` does not move.

Known gap, stated rather than papered over: `details_offered()` returns `false` while a
`persistence_warning` is up, which is precisely the reporter's case — so the on-screen half does
not reach the failure that prompted this. The logged per-candidate evidence does. Lifting that
suppression is a separate change.


GLinnik21:
Codex review, one round, on the combined result (this branch sits on 4f1c7874). One blocking finding, now fixed in 00aad70e.

**Both success paths discarded the accumulated diagnostics.** `save_legacy_fallback_locked` only logged `failures` on the total-failure paths; the sealed and plaintext loops each returned early on the first candidate that accepted the write. With 4f1c7874 on main that is no longer a rare shape — it is the common one: the durable candidates are refused and the runtime-directory candidate then succeeds, so the evidence explaining *why* /media/developer refused would have been thrown away in exactly the case this branch exists to explain. The diagnostics would only ever have surfaced when nothing worked at all.

Both paths now log the refusals before returning, through the same `log_candidate_diagnostics` formatter, with wording that says the save succeeded on a later candidate so the line cannot be misread as a failure. Silent when nothing was refused.

No finding on the other four: nothing secret reaches the log or the screen (paths, numeric ownership, errno — no token, server identity or username); the three screen fields were already serialized by `event_body`, so the PRIVACY.md wording is a correction and not a widening; every `write_atomic` call site preserves its previous boolean decision; and `storage_evidence_line` keeps three labelled slots for all eight presence combinations, including no context at all.

One factual note for the record, not a change here: `resolve_runtime_dir` returns `/tmp` itself for the stable app id, so a stable install's fallback is `/tmp/auth.json` in the shared root, while flavoured installs get `/tmp/<app id>/auth.json`. That is safe — 0600, sticky directory, owner check and O_NOFOLLOW — but the name is guessable in a shared namespace, so another app holding `auth.json` first would cost a stable install this fallback. Worth a follow-up alongside lifting the Details suppression.

